PR #3, opened 2018-07-13
Add a unit-test to check that V4 reads from PRNG fully

zerkms:
That's a test that demonstrates the original problem with V4 only being partially filled with random data

### DIFF
--- a/generator_test.go
+++ b/generator_test.go
@@ -22,6 +22,7 @@
 package uuid
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


theckman:
@zerkms I'll probably need to have you rebase this branch once #6 goes in. That way you can get a CI in before we merge. I'll give you a ping once that happens.

CameronAckermanSEL:
Are passing tests for this PR contingent on #7 ?

PR #8, opened 2018-07-13
fix package URLs in README.md

### LINKED ISSUE #1: Readme badges point to satori/go.uuid
Once we get Travis and similar things set up, we should make them point to this repository.

jadr2ddude:
Fixes #1 i think

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # UUID package for Go language
 
-[![Build Status](https://travis-ci.org/satori/go.uuid.svg?branch=master)](https://travis-ci.org/satori/go.uuid)
-[![Coverage Status](https://coveralls.io/repos/github/satori/go.uuid/badge.svg?branch=master)](https://coveralls.io/github/satori/go.uuid)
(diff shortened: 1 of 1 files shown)


CameronAckermanSEL:
Hey, we should also update the minimum supported version of Go from 1.2 to 1.3

jadr2ddude:
@CameronAckermanSEL done

README.md
```diff
@@ -19,11 +19,11 @@ Supported versions:
 
 Use the `go` command:
 
-	$ go get github.com/satori/go.uuid
+	$ go get github.com/gofrs/uuid
 
 ## Requirements
 
-UUID package requires Go >= 1.2.
+UUID package requires Go >= 1.3.
```
theckman: I don't think this is technically true. We stopped testing 1.2 because of our wanting to continue using coveralls.io, I'm not sure anything in the library itself is broken with `< Go 1.2`. Admittedly, because there is no context provided in da9da744cf7ea776632325cce5335dc49481529f someone may not know why we dropped Go 1.2 support.

We could maybe rephrase this to be more correct, in that this library is only **_tested_** against Go 1.3+.

jadr2ddude:
@theckman I have made the requested changes

PR #11, opened 2018-07-14
Rewrite default NewV2 to generate correct output if UID/GID are changed at runtime

jadr2ddude:
Previously, the package used the UID/GID from package initialization time. However the user may call syscall.Setuid at runtime.

While calling this at runtime is the correct behavior, it adds 2 syscalls per UUID created. This is a performance/correctness tradeoff.


jadr2ddude:
Closed due to https://github.com/golang/go/issues/1435

PR #12, opened 2018-07-14
documentation: add further clarification to comments and documentation

trevorstarick:


theckman:
@trevorstarick Thank you for updating the PR title. When you have a moment can you please amend your commit with @acln0's suggested change as well?

uuid.go
```diff
@@ -67,8 +68,7 @@ var (
 	byteGroups = []int{8, 4, 4, 4, 12}
 )
 
-// Nil is special form of UUID that is specified to have all
-// 128 bits set to zero.
+// Nil is special form of UUID that is specified to have all 128 bits set to zero.
```
theckman: I would refer back to the name used in the RFC:
>Nil is a nil UUID, as specified in RFC-4122, that has all 128 bits set to zero.

uuid.go
```diff
@@ -105,12 +105,12 @@ func (u UUID) Variant() byte {
...
 }
 
-// Bytes returns bytes slice representation of UUID.
+// Bytes returns a byte slice representation of the UUID.
 func (u UUID) Bytes() []byte {
 	return u[:]
 }
 
-// Returns canonical string representation of UUID:
+// String returns a canonical string representation of the UUID:
```
theckman: I'd add `RFC-4122` after the word canonical to make it more precise.

uuid.go
```diff
@@ -152,7 +152,7 @@ func (u *UUID) SetVariant(v byte) {
 // Must is a helper that wraps a call to a function returning (UUID, error)
 // and panics if the error is non-nil. It is intended for use in variable
 // initializations such as
-//	var packageUUID = uuid.Must(uuid.FromString("123e4567-e89b-12d3-a456-426655440000"));
```
theckman: >;

Hahahaha! Nice find! 😂

uuid.go
```diff
@@ -19,9 +19,11 @@
...
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-// Package uuid provides implementation of Universally Unique Identifier (UUID).
-// Supported versions are 1, 3, 4 and 5 (as specified in RFC 4122) and
-// version 2 (as specified in DCE 1.1).
+// Package uuid provides implementations of the Universally Unique Identifier (UUID), as specified in RFC 4122 and DCE 1.1.
+//
+// RFC 4122 (https://tools.ietf.org/html/rfc4122) provides the specification for versions 1, 3, 4, and 5.
+//
+// DCE 1.1 (http://pubs.opengroup.org/onlinepubs/9696989899/chap5.htm#tagcjh_08_02_01_01) provides the specification for version 2.
```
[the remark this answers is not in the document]
theckman: @trevorstarick actually, how do you feel about adding references to both?

```Go
// Package uuid provides implementations of the Universally Unique Identifier
// (UUID), as specified in RFC-4122[1] and DCE 1.1[2].
//
// RFC 4122 (https://tools.ietf.org/html/rfc4122) provides the specification for
// versions 1, 3, 4, and 5.
//
// DCE 1.1 provides the specification for version 2.
//
// [1] https://tools.ietf.org/html/rfc4122
// [2] http://pubs.opengroup.org/onlinepubs/9696989899/chap5.htm#tagcjh_08_02_01_01
```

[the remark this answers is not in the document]
trevorstarick: I [originally](http://starick.me/LyCQjb) did this, but found it weird to look at. Now that I've added more content I think it should look better.

Issue #15, opened 2018-07-14
Replace testing framework (go-check) with standard library assertions


theckman:
While overall there's nothing wrong with the `go-check` package, as least from my perspective as a user, it does add some overhead in to learning how to contribute unit tests to this package. We should look at replacing the go-check assertions with table-driven tests using the `testing` package from the stdlib.

(This is already a WIP)

PR #17, opened 2018-07-15
rework tests from gocheck to stdlib testing

acln0:
Use the standard library testing package instead of gocheck. Switch
from multiple assertions to table-driven tests where appropriate.

Update fuzz testing support: add code to generate a diverse corpus
based on existing parser tests, add instructions on how to run the
fuzzer, check in the existing corpus.


acln0:
There are probably things that I missed here. I'm sending this PR for preliminary review. It's fairly big, and it's going to need some eyes on it before a potential merge.

acln0:
Ugh. I completely forgot Go < 1.7 doesn't have subtests. Travis isn't happy. What should we do? I can get rid of all the subtests if we need to.

theckman:
@acln0 I wouldn't be against dropping the test runners for the older Go versions within the context of this PR.

Outside of this PR, we should have the discussion as an organization as to whether our support of Go runtime versions differs from the authors.

acln0:
Filed #18 for discussion.

theckman:
@acln0 can you rebase this branch against `master` and then force-push it back up? That should get this to be green I think.

uuid_test.go
```diff
@@ -25,76 +25,120 @@ import (
...
+		{
+			u:    UUID{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00},
+			want: VariantRFC4122,
+		},
+		{
+			u:    UUID{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xc0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00},
+			want: VariantMicrosoft,
+		},
+		{
+			u:    UUID{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xe0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00},
```
jadr2ddude: Does this work? If so, how?

jadr2ddude: Nevermind i misread

acln0:
@theckman Done.

@jadr2ddude Yes, that works, because the `Variant` method only touches that single byte.

theckman:
@acln0 awesome! I did a quick pass on this PR before, but let me give it a look again tonight. It'll be a few hours until I'm at home and able to give it some proper :eyes:.

acln0:
Thanks. I forgot to mention it in the commit message, but, once merged, this closes #2 and #15.

PR #21, opened 2018-07-16
Disable testing against Go 1.3 ~ 1.6 to improve tests

### LINKED ISSUE #18: Minimum supported Go version
This matter came to light with the submission of #17. In reworking the tests, I unfortunately omitted the fact that Go < 1.7 doesn't have subtests, so the Travis builds for < 1.7 are failing.

In the context of that PR, I think the relevant questions are whether or not testing for Go >= 1.3 && Go < 1.7 adds value for users who are stuck on those versions, and whether they'd be comfortable using the package if regular tests only run on Go >= 1.7.

If we assume that the package will have users in that category, and we believe the answer to the former question to be "yes" (and the answer to the latter to be "no"), then I can refactor the tests to work on < 1.7.

On the other hand, subtests are a nice modernization, and they fit the nature of the tests quite well.

If the subtests stay, the documentation w.r.t. supported versions could read "tested on >= 1.7, probably works on >= 1.2". If we remove subtests, the documentation will continue to say "tested on >= 1.3, probably works on >= 1.2".

Thoughts?

theckman:
In preparation for executing on #15 (already opened in #17) we need to disable
testing against older versions of the Go toolchain. There are a few reasons for
this, the largest being they don't support subtests. These versions, including
Go 1.7 and 1.8, are EOL by the Go authors so there's also lack of upstream
support for these toolchains.

This was discussed a bit in #18 and on [Slack](https://gophers.slack.com/archives/CBP4N9BEU/p1531704009000014).

Fixes #18


theckman:
>0 of 0 changed or added relevant lines in 0 files are covered.
>Overall coverage decreased (-0.006%) to 98.131%
@coveralls wat?

README.md
```diff
@@ -49,9 +49,10 @@ $ go get github.com/gofrs/uuid
 
 ## Requirements
 
-This package is regularly tested against Go 1.3+, due to issues with the testing
-libraries on Go 1.2 or less. This package may work perfectly fine with Go 1.2,
-but it is not tested and support for this version is not actively maintained.
+This package is regularly tested against Go 1.7+, due to older versions not
+supporting subtests. This package may work perfectly fine with Go 1.2+, but it
+is not tested against and support for these older versions is not actively
```
acln0: > but it is not tested against and support for these [...]

This phrasing sounds a little odd to me. It may be because I'm not a native English speaker, but I had to read it a couple of times to understand the meaning.

theckman:
Gonna trigger a CI re-run.

README.md
theckman: @acln0 patch applied -- how's it look now?

PR #22, opened 2018-07-16
Added couple tests that increase code test coverage

zerkms:
The `TestDecodePlainWithWrongLength` runs unexported method since there is no other way to reach that line.


zerkms:
WAIT WHAT

PR #24, opened 2018-07-17
Covered UUID.decodePlain() with wrong input with a test

zerkms:
The `TestDecodePlainWithWrongLength` runs unexported method since there is no other way to reach that line.

It's a substitution for the https://github.com/gofrs/uuid/pull/22 which I accidentally closed via deleting my branch (instead of hard-push'ing it)

### DIFF
--- a/codec_test.go
+++ b/codec_test.go
@@ -222,6 +222,16 @@ func TestMarshalText(t *testing.T) {
 	}
 }
 
+func TestDecodePlainWithWrongLength(t *testing.T) {
+	arg := []byte{'4', '2'}
+
+	u := UUID{}
+
+	if u.decodePlain(arg) == nil {
+		t.Errorf("%v.decodePlain(%q): should return error, but it did not", u, arg)
+	}
+}
+
 var stringBenchmarkSink string
 
 func BenchmarkString(b *testing.B) {


theckman:
@zerkms I just destroyed your branch, I'm sorry. :(

zerkms:
It's fine, I need to include a tiny change there anyway :-)

codec_test.go
```diff
@@ -222,6 +222,17 @@ func TestMarshalText(t *testing.T) {
 	}
 }
 
+func TestDecodePlainWithWrongLength(t *testing.T) {
+	arg := []byte{'4', '2'}
+
+	u := UUID{}
+	err := u.decodePlain(arg)
```
theckman: @zerkms I might be nitpicking here a bit. Since we do not use the error value, how do we feel about simplifying this to the single if expression?

```Go
if u.decodePlain(arg) == nil {}
```

theckman: The alternative I'd go with, if we care about the error type, would be:

```Go
var err error
// ...
if err = u.decodePlain(arg); err == nil {}
```

Just so that we assert the return type as being an `error` value.

zerkms: I don't mind in either, but would personally prefer the first (given every other test case does not care about the exact returned `err` type either).

PR #28, opened 2018-07-18
update code example in readme

acln0:
Make the example code compile, demonstrate the more common use case
for Must, improve example program logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,30 +62,30 @@ documentation, please see the [GoDoc Page](http://godoc.org/github.com/gofrs/uui
 package main
 
 import (
-	"fmt"
+	"log"
+
 	"github.com/gofrs/uuid"
 )
 
-func main() {
-	// Creating UUID Version 4
-	// panic on error
-	u1 := uuid.Must(uuid.NewV4())
-	fmt.Printf("UUIDv4: %s\n", u1)
+// Create a Version 4 UUID, panicking on error.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,30 +62,30 @@ documentation, please see the [GoDoc Page](http://godoc.org/github.com/gofrs/uui
...
+	log.Printf("generated Version 4 UUID %v", u2)
 
-	// Parsing UUID from string input
-	u2, err := uuid.FromString("6ba7b810-9dad-11d1-80b4-00c04fd430c8")
+	// Parse a UUID from a string.
+	s := "6ba7b810-9dad-11d1-80b4-00c04fd430c8"
+	u3, err := uuid.FromString(s)
 	if err != nil {
-		fmt.Printf("Something went wrong: %s", err)
+		log.Fatalf("failed to parse UUID %q: %v", s, err)
```
theckman: Why `%v` instead of `%s`? I think we'd want the error message here, which would be `%s`.

acln0: I'm not sure `%s` for errors is right. The documentation for `%s` says "the uninterpreted bytes of the string or slice". AFAIK, `%v` for errors is interpreted as the result of `err.Error()`.

theckman: Looks like `%v` and `%s` behave the same. TIL. 👍

PR #29, opened 2018-07-18
Update package comments to provide better documentation

### LINKED ISSUE #27: Possible Godoc improvements
I think https://godoc.org/github.com/gofrs/uuid#Equal should say "Nil is the nil UUID...".

https://godoc.org/github.com/gofrs/uuid#UUID.UnmarshalText should mention the braced hashlike format, as described in #14, and the URN hashlike format as in https://github.com/gofrs/uuid/blob/57be1465e1906c395f54925bc002ce4428732410/codec_test.go#L131-L134.

Perhaps https://godoc.org/github.com/gofrs/uuid#DefaultGenerator could be placed outside the `var` block it is currently in, standing alone instead, to reduce noise and to promote the comment to a "proper Godoc" rather than code comment.

Perhaps the comment on https://godoc.org/github.com/gofrs/uuid#UUID could also be improved. It feels like it needs a verb.

theckman:
This updates the package comments in different areas to improve the GoDoc
rendering of the documentation.

Fixes #27

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -71,27 +71,38 @@ func (u UUID) MarshalText() ([]byte, error) {
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
 // Following formats are supported:
+//
 //   "6ba7b810-9dad-11d1-80b4-00c04fd430c8",
 //   "{6ba7b810-9dad-11d1-80b4-00c04fd430c8}",
 //   "urn:uuid:6ba7b810-9dad-11d1-80b4-00c04fd430c8"
 //   "6ba7b8109dad11d180b400c04fd430c8"
+//   "{6ba7b8109dad11d180b400c04fd430c8}",
+//   "urn:uuid:6ba7b8109dad11d180b400c04fd430c8"
+//
 // ABNF for supported UUID text representation follows:
-//   uuid := canonical | hashlike | braced | urn
-//   plain := canonical | hashlike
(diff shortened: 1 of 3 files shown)


codec.go
```diff
@@ -71,27 +71,38 @@ func (u UUID) MarshalText() ([]byte, error) {
...
+//
+//   hashlike := 12hexoct
+//   canonical := 4hexoct '-' 2hexoct '-' 2hexoct '-' 6hexoct
+//
+//   plain := canonical | hashlike
+//   uuid := canonical | hashlike | braced | urn
+//
+//   braced := '{' plain '}' | '{' hashlike  '}'
+//   urn := URN ':' UUID-NID ':' plain
+//
```
acln0: I think this newline can be removed.

theckman: @acln0 I intentionally kept it because it otherwise looked too bunched-up next to the function header, when viewing the source code. This is a pretty large comment, so it helped.

screen shot 2018-07-18 at 12 43 02 am

acln0: Aha, I see. Well, if it looks fine in Godoc, we can keep it. That was the only concern.

theckman: To loop back around, looks to be okay!
screen shot 2018-07-18 at 12 46 47 am

acln0: Great!

theckman:
🖕@coveralls

PR #31, opened 2018-07-18
Add Time() and Nanos() methods

rkuris:
These methods can be used to extract time values
from V1 UUIDs

Originally added as https://github.com/satori/go.uuid/pull/50

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -30,10 +30,10 @@
 package uuid
 
 import (
-	"encoding/hex"
 	"encoding/binary"
-	"time"
+	"encoding/hex"
 	"fmt"
+	"time"
 )
 
 // Size of a UUID in bytes.
@@ -76,8 +76,8 @@ const _100nsPerSecond = 10000000
 
 // Time returns the UTC time.Time representation of a Timestamp
 func (t Timestamp) Time() (time.Time, error) {
-	secs := uint64(t)/_100nsPerSecond
-	nsecs := 100 * (uint64(t)%_100nsPerSecond)
+	secs := uint64(t) / _100nsPerSecond
+	nsecs := 100 * (uint64(t) % _100nsPerSecond)
 	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)), nil
 }
 
@@ -91,7 +91,7 @@ func TimestampFromV1(u UUID) (Timestamp, error) {
 	low := binary.BigEndian.Uint32(u[0:4])
 	mid := binary.BigEndian.Uint16(u[4:6])
 	hi := binary.BigEndian.Uint16(u[6:8]) & 0xfff
-	return Timestamp(low) + (Timestamp(mid) << 32) + (Timestamp(hi) << 48), nil
+	return Timestamp(uint64(low) + (uint64(mid) << 32) + (uint64(hi) << 48)), nil
 }
 
 // String parse helpers.
--- a/uuid.go
+++ b/uuid.go
@@ -31,6 +31,9 @@ package uuid
 
 import (
 	"encoding/hex"
+	"encoding/binary"
+	"time"
+	"fmt"
 )
 
 // Size of a UUID in bytes.
@@ -63,6 +66,33 @@ const (
 	DomainGroup
 	DomainOrg
 )
+
+// Timestamp is the count of 100-nanosecond intervals since 00:00:00.00,
+// 15 October 1582 within a V1 UUID. This type has no meaning for V2-V5
+// UUIDs since they don't have an embedded timestamp.
+type Timestamp uint64
+
+const _100nsPerSecond = 10000000
+
+// Time returns the UTC time.Time representation of a Timestamp
+func (t Timestamp) Time() (time.Time, error) {
+	secs := uint64(t)/_100nsPerSecond
+	nsecs := 100 * (uint64(t)%_100nsPerSecond)
+	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)), nil
+}
+
+// TimestampFromV1 returns the Timestamp embedded within a V1 UUID.
+// Returns an error if the UUID is any version other than 1.
+func TimestampFromV1(u UUID) (Timestamp, error) {
+	if u.Version() != 1 {
+		err := fmt.Errorf("uuid: %s is version %d, not version 1", u, u.Version())
+		return 0, err
+	}
+	low := binary.BigEndian.Uint32(u[0:4])
+	mid := binary.BigEndian.Uint16(u[4:6])
+	hi := binary.BigEndian.Uint16(u[6:8]) & 0xfff
+	return Timestamp(low) + (Timestamp(mid) << 32) + (Timestamp(hi) << 48), nil
+}
 
 // String parse helpers.
 var (
--- a/uuid_test.go
+++ b/uuid_test.go
@@ -137,21 +137,21 @@ func TestMust(t *testing.T) {
 
 func TestTimeFromTimestamp(t *testing.T) {
 	tests := []struct {
-		t Timestamp
+		t    Timestamp
 		want time.Time
 	}{
 		// a zero timestamp represents October 15, 1582 at midnight UTC
-		{ t: Timestamp(0), want: time.Date(1582, 10, 15, 0, 0, 0, 0, time.UTC) },
+		{t: Timestamp(0), want: time.Date(1582, 10, 15, 0, 0, 0, 0, time.UTC)},
 		// a one value is 100ns later
-		{ t: Timestamp(1), want: time.Date(1582, 10, 15, 0, 0, 0, 100, time.UTC) },
+		{t: Timestamp(1), want: time.Date(1582, 10, 15, 0, 0, 0, 100, time.UTC)},
 		// 10 million 100ns intervals later is one second
-		{ t: Timestamp(10000000), want: time.Date(1582, 10, 15, 0, 0, 1, 0, time.UTC) },
-		{ t: Timestamp(60*10000000), want: time.Date(1582, 10, 15, 0, 1, 0, 0, time.UTC) },
-		{ t: Timestamp(60*60*10000000), want: time.Date(1582, 10, 15, 1, 0, 0, 0, time.UTC) },
-		{ t: Timestamp(24*60*60*10000000), want: time.Date(1582, 10, 16, 0, 0, 0, 0, time.UTC) },
-		{ t: Timestamp(365*24*60*60*10000000), want: time.Date(1583, 10, 15, 0, 0, 0, 0, time.UTC) },
+		{t: Timestamp(10000000), want: time.Date(1582, 10, 15, 0, 0, 1, 0, time.UTC)},
+		{t: Timestamp(60 * 10000000), want: time.Date(1582, 10, 15, 0, 1, 0, 0, time.UTC)},
+		{t: Timestamp(60 * 60 * 10000000), want: time.Date(1582, 10, 15, 1, 0, 0, 0, time.UTC)},
+		{t: Timestamp(24 * 60 * 60 * 10000000), want: time.Date(1582, 10, 16, 0, 0, 0, 0, time.UTC)},
+		{t: Timestamp(365 * 24 * 60 * 60 * 10000000), want: time.Date(1583, 10, 15, 0, 0, 0, 0, time.UTC)},
 		// maximum timestamp value in a UUID is the year 5236
-		{ t: Timestamp(uint64(1<<60 - 1)), want: time.Date(5236, 03, 31, 21, 21, 0, 684697500, time.UTC) },
+		{t: Timestamp(uint64(1<<60 - 1)), want: time.Date(5236, 03, 31, 21, 21, 0, 684697500, time.UTC)},
 	}
 	for _, tt := range tests {
 		got, _ := tt.t.Time()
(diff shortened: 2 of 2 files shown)


rkuris:
I'd gladly stop using my fork if this can be maintained by gofrs; suggest bumping to revision 2.1.0 when adding this functionality.

uuid.go
```diff
@@ -153,6 +156,28 @@ func (u *UUID) SetVariant(v byte) {
 	}
 }
 
+// Nanos returns the number of nanoseconds in a V1 UUID
+func (u *UUID) Nanos() (int64, error) {
```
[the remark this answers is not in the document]
rkuris: Good suggestion. Completed.

uuid.go
```diff
@@ -153,6 +156,28 @@ func (u *UUID) SetVariant(v byte) {
...
+	}
+	low := binary.BigEndian.Uint32(u[0:4])
+	mid := binary.BigEndian.Uint16(u[4:6])
+	hi := binary.BigEndian.Uint16(u[6:8]) & 0xfff
+	return 100 * (int64(low) + (int64(mid) << 32) + (int64(hi) << 48) - int64(epochStart)), nil
+}
+
+// Time returns the time embedded in a V1 UUID
+// An error is returned if this is not a V1 UUID
+func (u *UUID) Time() (time.Time, error) {
```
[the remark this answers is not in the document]
rkuris: Done.

uuid.go
```diff
@@ -153,6 +156,28 @@ func (u *UUID) SetVariant(v byte) {
 	}
 }
 
+// UnixNanoOf returns the number of nanoseconds in a V1 UUID
```
theckman: Just to be very clear to consumers, would you mind adding a quip about it being the number of nanoseconds since the UNIX epoch? I definitely don't want to nitpick copy, but I'm thinking something like:

>// UnixNanoOf returns the number of nanoseconds elapsed since January 1, 1970 UTC in a V1 UUID

It might also be good to indicate that an error is returned for a non-V1 UUID here too (like we did with `TimeOf`).

Finally, please add punctuation to the sentence just so that the documentation reads better.

uuid.go
```diff
@@ -153,6 +156,28 @@ func (u *UUID) SetVariant(v byte) {
...
+func (u *UUID) UnixNanoOf() (int64, error) {
+	if u.Version() != 1 {
+		err := fmt.Errorf("uuid: %s is version %d, not version 1", u, u.Version())
+		return 0, err
+	}
+	low := binary.BigEndian.Uint32(u[0:4])
+	mid := binary.BigEndian.Uint16(u[4:6])
+	hi := binary.BigEndian.Uint16(u[6:8]) & 0xfff
+	return 100 * (int64(low) + (int64(mid) << 32) + (int64(hi) << 48) - int64(epochStart)), nil
```
theckman: When you have a moment would you please add some spacing to this block of code and put the return calculation on a separate line (for readability purposes)? For spacing I'm just thinking of an empty line before the `low` line and after the `hi` line.

For the return, I think it'd be okay to put the calculation on its own line so that we can then have `return var, nil`. For me at least, it made it easier to grok on a quick look.

uuid.go
```diff
@@ -153,6 +156,28 @@ func (u *UUID) SetVariant(v byte) {
...
+}
+
+// TimeOf returns the time embedded in a V1 UUID
+// An error is returned if this is not a V1 UUID
+func (u *UUID) TimeOf() (time.Time, error) {
```
theckman: Would you mind converting this to a function on the package that takes the UUID as input? Because this method only takes a UUID, I think it's more appropriate to be a function. If the UUID wasn't the input to the method, but was the state needed for the method to take action on the input, then I'd say 👍 on it being a method.

uuid.go
```diff
@@ -153,6 +156,28 @@ func (u *UUID) SetVariant(v byte) {
 	}
 }
 
+// UnixNanoOf returns the number of nanoseconds in a V1 UUID
+func (u *UUID) UnixNanoOf() (int64, error) {
```
theckman: Would you be able to make this a package function too?

uuid.go
```diff
@@ -153,6 +156,28 @@ func (u *UUID) SetVariant(v byte) {
...
+		err := fmt.Errorf("uuid: %s is version %d, not version 1", u, u.Version())
+		return 0, err
+	}
+	low := binary.BigEndian.Uint32(u[0:4])
+	mid := binary.BigEndian.Uint16(u[4:6])
+	hi := binary.BigEndian.Uint16(u[6:8]) & 0xfff
+	return 100 * (int64(low) + (int64(mid) << 32) + (int64(hi) << 48) - int64(epochStart)), nil
+}
+
+// TimeOf returns the time embedded in a V1 UUID
```
theckman: This may seem pedantic, but would you be able to add punctuation after `UUID` on this line? I'll make the GoDoc read better.

generator_test.go
```diff
@@ -142,10 +156,53 @@ func testNewV1MissingNetworkFaultyRand(t *testing.T) {
...
+}
+
 func testNewV2(t *testing.T) {
 	t.Run("Basic", testNewV2Basic)
 	t.Run("DifferentAcrossCalls", testNewV2DifferentAcrossCalls)
 	t.Run("FaultyRand", testNewV2FaultyRand)
+	t.Run("TestTimeOfErrors", testTimeOfErrors)
+}
+
+func testTimeOfErrors(t *testing.T) {
```
theckman: Can we also add a test that would use a constant UUID, and test to make sure that `UnixNanosOf()` and `TimeOf()` parse the time out of it properly?

generator_test.go
```diff
@@ -45,6 +45,7 @@ func testNewV1(t *testing.T) {
 	t.Run("FaultyRand", testNewV1FaultyRand)
 	t.Run("MissingNetwork", testNewV1MissingNetwork)
 	t.Run("MissingNetworkFaultyRand", testNewV1MissingNetworkFaultyRand)
+	t.Run("V1Ordering", testNewV1Ordering)
```
acln0: Please change `"V1Ordering"` to `"Ordering"`, since this is a subtest of `testNewV1`.

generator_test.go
```diff
@@ -142,10 +156,53 @@ func testNewV1MissingNetworkFaultyRand(t *testing.T) {
...
+	n1, err := u1.UnixNanoOf()
+	if err != nil {
+		t.Fatal(err)
+	}
+	n2, err := u2.UnixNanoOf()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if n1 > n2 {
+		t.Errorf("generated older UUID: %s > %s", u1, u2)
```
acln0: Perhaps this message should print the nanoseconds, like the one below prints time.

generator_test.go
acln0: This test does not test a property of `NewV2`, so I don't think it should be a subtest of `testNewV2`.

zerkms:
After giving it some more thoughts and reading the spec I came with the following new types to introduce

```go
// 100s of nanos since 00:00:00.00, 15 October 1582
type V1Timestamp int64

func TimestampFromV1(UUID) (V1Timestamp, error)

func TimeFromV1Timestamp(V1Timestamp) (time.Time, error)
```

Rationale behind it:

* Those are functions not methods of `UUID` since they make sense only for V1 and not the other types of UUID

* The `V1Timestamp` type specifically chosen according the RFC terminology (https://tools.ietf.org/html/rfc4122#section-4.1.4)

* `V1Timestamp` values are comparable by themselves, if one needs exact time - there is a converting function for convenience

acln0:
@zerkms I like this API very much. It's a little verbose, but it's very explicit, and it encodes the notion that timestamps are in 100s of nanoseconds since the Gregorian calendar in a type. Big +1 from me.

jadr2ddude:
Would it be a better idea to do V1Timestamp.Time() or TimeFromV1Timestamp(V1Timestamp) ?

acln0:
I have no opinion on the method vs. top-level function matter, but I do like the proposed data type, and I think we should use it.

rkuris:
+1 from me on the new type. I will work on the proposed changes. Since
there is a lot of indifference in the method vs function debate, how about
we just offer both?

jadr2ddude:
@rkuris it doesn't really make sense to do both. I would be fine with either.

theckman:
I would prefer we only offer a function, and not both.

acln0:
I agree with not offering both. Let's have one, obvious way to do things.

That being said, an argument can be made for the method variant, based on prior cases. Note, for example, that the `time` package offers https://golang.org/pkg/time/#Time.Minute and https://golang.org/pkg/time/#Time.Month rather than `MinuteFromTime` and `MonthFromTime`.

I wouldn't be opposed to an API like

`func (u UUID) V1Timestamp() (V1Timestamp, error)`

and

`func (t V1Timestamp) Time() (time.Time, error)`

Both this and the function-based API read well to me. I suppose it's a question of taste and style. We need a resolution nevertheless, so we can move forward with the proposal.

theckman:
I think I'm hesitant to have it on the type because not every `UUID` is a V1 UUID.

acln0:
@theckman Thanks. That's a very good point, and I am now leaning strongly towards the function variant as well.

theckman:
We've had quite a few discussions in this PR, so I think it'd be good to summarize what we've talked about.

I think the agreement is that we want to create a new unsigned integer timestamp type [1], to represent the number of 100-nanosecond intervals since 00:00:00.00, 15 October 1582 [2]. This is to be closer to the RFC and how it thinks about time. A user will obtain one of these timestamps by calling a function on the _package_, not a method on a type within the package, while giving it a V1 UUID which it can parse the time out of.

To then get a `time.Time` value, I'm of the opinion that it should be a method on this timestamp type (taking inspiration from the `time` package).

I like the idea of the proposal so far, but I've been thinking and I want to propose a slight tweak and see what people think. Based on reading the RFC again, the "timestamp" of V2 ~ V5 UUIDs isn't actually a timestamp at all, confirming that only applies to V1. So with this I think specifying `V1` is a bit redundant. We can just mention it in the docs for those who are familiar with the RFC.

Here's what I was thinking of, to expand on the example above:

```Go
// 15 October 1582 within a V1 UUID. This type has no meaning for
// V2 ~ V5 UUIDs as they don't have an embedded timestamp .
type Timestamp uint64

// Time returns the UTC time.Time representation of the Timestamp.
func (t Timestamp) Time() (time.Time, error)

// TimestampFromV1 returns the Timestamp embedded within a V1 UUID.
// Returns an error if the UUID is any version other than 1.
func TimestampFromV1(u UUID) (Timestamp, error)
```

What does everyone think? Would this be acceptable for everyone as the API we want to support?

[1] https://tools.ietf.org/html/rfc4122#section-4.2.2
[2] https://tools.ietf.org/html/rfc4122#section-4.1.4

theckman:
@rkuris let us know if there's anything we can do to help with this PR.

rkuris:
Should be able to code this up this week. Thanks for all the great input!

theckman:
@rkuris I wanted to follow back up and see if there's anything we can do to help you with this PR. I think we'd be happy to adapt your work and raise a separate PR, if you're okay with that. If we do go that route, I'd like to do it in a way that tries to make sure we are able to still attribute the contribution to you in some capacity.

rkuris:
Sorry -- I had some work issues pop up. If I can't complete it this week,
then by all means, take over and implement it. I don't think it's a lot of
work.

Ron

rkuris:
Testing the edge cases turned up a bug in how times were constructed in the previous PR. I'll make a note in the satori.uuid repo to that effect.

rkuris:
Greetings,

I think all the comments have been addressed by this rewrite. Is this ready to merge now or does this need more work?

uuid.go
```diff
@@ -64,6 +67,33 @@ const (
...
 
+// Timestamp is the count of 100-nanosecond intervals since 00:00:00.00,
+// 15 October 1582 within a V1 UUID. This type has no meaning for V2-V5
+// UUIDs since they don't have an embedded timestamp.
+type Timestamp uint64
+
+const _100nsPerSecond = 10000000
+
+// Time returns the UTC time.Time representation of a Timestamp
+func (t Timestamp) Time() (time.Time, error) {
```
zerkms: Is `error` as a second return value here for potential unknown cases in future to avoid BCs, or is there any other reason?

I'm not against, or pro- having it, just curious.

uuid.go
```diff
@@ -64,6 +67,33 @@ const (
...
+// Returns an error if the UUID is any version other than 1.
+func TimestampFromV1(u UUID) (Timestamp, error) {
+	if u.Version() != 1 {
+		err := fmt.Errorf("uuid: %s is version %d, not version 1", u, u.Version())
+		return 0, err
+	}
+	low := binary.BigEndian.Uint32(u[0:4])
+	mid := binary.BigEndian.Uint16(u[4:6])
+	hi := binary.BigEndian.Uint16(u[6:8]) & 0xfff
+	return Timestamp(low) + (Timestamp(mid) << 32) + (Timestamp(hi) << 48), nil
```
zerkms: I think using `Timestamp` type here is confusing.

How about doing math + bit shift operations with `uint64` first?

uuid.go
rkuris: I think that's the idea. You can use Must(thing.Time()) if you don't care about errors.

uuid.go
rkuris: A Timestamp is a uint64, and we need to return one. Why is it confusing?

uuid_test.go
```diff
@@ -133,3 +134,51 @@ func TestMust(t *testing.T) {
 	}
 	Must(fn())
 }
+
+func TestTimeFromTimestamp(t *testing.T) {
+	tests := []struct {
```
acln0: `gofmt` please.

uuid.go
zerkms: I'm a proponent of coding against types: I strongly believe that types carry a lot of extra context.

In this case, keeping in mind that `Timestamp` is a type that represents number of 100s of nanoseconds since `00:00:00.00 15 October 1582`, the `Timestamp(mid) << 32` operation makes very little sense: what does "shift 100s of nanoseconds 32 bits left" mean?

It's a mathematical operation and should be applied to numbers, which then (in the very end) might be interpreted as something domain-specific.

PS: it's my personal opinion, if other people are okay - I won't insist on this change too hard.

acln0:
LGTM now.

uuid.go
rkuris: I changed it. I don't have a strong opinion either way, but to me it seems harder to read after making the change (a ton of casting is now happening).

I suppose I could break it up into multiple lines but it still seems "wordier". I don't have a strong opinion though.

zerkms: It looks ideal to me now :-)

theckman:
I'll review / merge tomorrow morning (in ~8 hours).

theckman:
New release cut that includes this change: https://github.com/gofrs/uuid/releases/tag/v3.1.0

PR #34, opened 2018-07-19
Improve Equal package function

### LINKED ISSUE #33: Questionable documentation and implementation of Equals
The documentation for https://godoc.org/github.com/gofrs/uuid#Equal could be improved. While we're at it, perhaps the implementation could simply `return u1 == u2`.

theckman:
This change makes some improvements to the Equal package function:

- updates function comment to read better
- changes the argument variables to `a` and `b`, instead of `u1` and `u2`
- compare the arrays directly, instead of converting to slices and comparing

Fixes #33

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -30,7 +30,6 @@
 package uuid
 
 import (
-	"bytes"
 	"encoding/hex"
 )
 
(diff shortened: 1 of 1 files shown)


acln0:
What happened with the 1.7 build?

theckman:
@acln0 looks like the runner ate it. Re-running that one.

PR #36, opened 2018-07-19
Deprecate the Equal package function

### LINKED ISSUE #35: Deprecate Equal package function
Because two values of a `UUID` (array) type are directly comparable, the `Equal()` function is superfluous. We should add a deprecation notice to the function comment, with the intent to remove it in v3.0.0.

theckman:
This deprecates the `Equal` package function, in favor of consumers using the
`==` comparison operator themselves between two `UUID` values. Because the UUID
value is an array type, there is no reason for the helper function.

The intent is to remove this as part of a v2.x => v3.x release.

Fixes #35

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -83,6 +83,9 @@ var (
 )
 
 // Equal returns true if a and b are equivalent.
+//
+// Deprecated: this function is deprecated and will be removed in a future major
+// version, as values of type UUID are directly comparable using `==`.
 func Equal(a UUID, b UUID) bool {
 	return a == b
 }


uuid.go
```diff
@@ -83,6 +83,10 @@ var (
 )
 
 // Equal returns true if a and b are equivalent.
+//
+// Deprecated: this function is deprecated and will be removed in a future major
+// version. The UUID type itself is directly comparable to another UUID value
+// using `==`, meaning this function superfluous.
```
acln0: "meaning this function superfluous" does not sound quite right.

acln0: Perhaps more of the phrase can be simplified by saying "values of type UUID are comparable using ==".

theckman:
I'm going to be traveling over the coming week, leaving tomorrow morning (~9 hours). I'll try to either get the commit amended in the airport, or after I get to my hotel.

theckman:
Commit amended and pushed.

Issue #32, opened 2018-07-19
Outstanding open issues / PRs in satori/go.uuid


acln0:
We should take a look at the open issues and pending PRs in https://github.com/satori/go.uuid and see if which ones would be worth tackling / adopting.

https://github.com/satori/go.uuid/pull/44 was brought up, which I think is a good start.

Please comment with other issues or PRs if you think they'd be worth doing.

theckman:
I pinged PR-44 to ask if they'd be willing to re-open the PR here.

theckman:
So here's a quick classification of their issues and PRs:

Desired / In-Progress
- https://github.com/satori/go.uuid/pull/44 (https://github.com/gofrs/uuid/pull/38)
- https://github.com/satori/go.uuid/pull/50 (https://github.com/gofrs/uuid/pull/31)

Fixed by this package:
- https://github.com/satori/go.uuid/pull/72
- https://github.com/satori/go.uuid/pull/75
- https://github.com/satori/go.uuid/pull/85

Obsoleted by this package
- https://github.com/satori/go.uuid/pull/68 (in that we've released a major version)

Unsure if we want to support this:
- https://github.com/satori/go.uuid/pull/63 (Request: https://github.com/satori/go.uuid/issues/64)

**_Issues_**

Feature requests:
- https://github.com/satori/go.uuid/issues/64 (PR: https://github.com/satori/go.uuid/pull/63)
- https://github.com/satori/go.uuid/issues/40

Deprecation requests:
- https://github.com/satori/go.uuid/issues/84
- https://github.com/satori/go.uuid/issues/83

Fixed by this package:
- https://github.com/satori/go.uuid/issues/82
- https://github.com/satori/go.uuid/issues/80
- https://github.com/satori/go.uuid/issues/77
- https://github.com/satori/go.uuid/issues/76
- https://github.com/satori/go.uuid/issues/73
- https://github.com/satori/go.uuid/issues/71
- https://github.com/satori/go.uuid/issues/70
- https://github.com/satori/go.uuid/issues/69
- https://github.com/satori/go.uuid/issues/66

Random questions (should probably be closed):
- https://github.com/satori/go.uuid/issues/81
- https://github.com/satori/go.uuid/issues/79
- https://github.com/satori/go.uuid/issues/78
- https://github.com/satori/go.uuid/issues/74
- https://github.com/satori/go.uuid/issues/47
- https://github.com/satori/go.uuid/issues/38

theckman:
[...]

PR #38, opened 2018-07-23
Add JSON (Un)marshaling methods on NullUUID type

theckman:
This commit adapts satori/go.uuid#44 for our fork of the
original project. This brings JSON Marshaling and Unmarshaling to the
`uuid.NullUUID` type used for use with SQL databases.

This is needed because `uuid.NullUUID` is a shim around `uuid.UUID`, providing
the information needed by the `database/sql` package to support storing the
value in a nullable column. Without these methods, the type gets converted
unnecessarily to a different format than the standard `uuid.UUID` type.

### DIFF
--- a/sql.go
+++ b/sql.go
@@ -22,7 +22,9 @@
 package uuid
 
 import (
+	"bytes"
 	"database/sql/driver"
+	"encoding/json"
 	"fmt"
 )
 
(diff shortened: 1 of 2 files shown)


theckman:
This is part of #32.

theckman:
This is going to actually warrant a 3.0.0. I think changing how this serializes to JSON could be a breaking change.

PaluMacil:
Should summary issue #32 be updated to mark this as merged?

philippfrank:
Came here late from the original repo, thanks for taking over maintenance. Out of curiosity, why did you put the JSON marshaller into `sql.go`?

PR #42, opened 2018-07-28
Expose the reference UUID generator constructor

### LINKED ISSUE #41: Add support for custom Hardware (MAC) Addresses
Some consumers of this package may want to generate `V1` UUIDs without exposing their physical device's hardware address to consumers of that UUID. We should look at exposing this functionality in some capacity for consumers.

See the issue on the original repo for more details (https://github.com/satori/go.uuid/issues/64).

/cc @alessandrozucca

It seems like exposing the generator itself, and allowing people to provide their own HWAddr function should support most of what's needed here.

theckman:
This change exposes the reference UUID generator in the package, so that
consumers can create a new instance and make use of it. This is being done so
that consumers can provide their own Hardware Address generation function, in
case they want to randomize the MAC address as to not expose it.

This works by now exposing the `rfc4122Generator` type as `Gen`, and adding
constructor functions that scaffold the generator appropriately. The tests were
also enhanced to confirm that the `HWAddrFunc` we provide is indeed being
invoked.

Fixes #41


generator.go
```diff
@@ -84,31 +86,61 @@ type Generator interface {
...
+//
+// This is used so that consumers can generate their own MAC addresses, for use
+// in the generated UUIDs, if there is some concern about exposing the physical
+// address of the machine generating the UUID.
+//
+// The Gen generator will only invoke the HWAddrFunc once, and cache that MAC
+// address for all the future UUIDs generated by it. If you'd like to switch the
+// MAC address being used, you'll need to create a new generator using this
+// function.
+func NewGenHWAF(hwaf HWAddrFunc) *Gen {
```
theckman: I'll be the first to admit that I'm not sold on this function name. Couldn't think of anything better in the moment, though.

zerkms: I personally tend to name things with `With` in such cases: `NewGenWithHWAF`

theckman: That sounds better to me.

zerkms:
@jadr2ddude having it as a factory function abstracts UUID client from its constructor: you as a client may generate new UUIDs and be isolated from making a decision on how and where from the hardware address is obtained from.

theckman:
@jadr2ddude I am assuming less than 10% of users will make use of this functionality, and so I don't think it makes sense to require the 90% to consider their hardware address as part of the API they interact with. For the majority we should silently do what they expect.

PR #54, opened 2018-09-11
Tiny code simplification in codec

ferhatelmas:
Drop unnecessary slice since type is already a slice.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -125,7 +125,7 @@ func (u *UUID) decodeCanonical(t []byte) error {
 		return fmt.Errorf("uuid: incorrect UUID format %s", t)
 	}
 
-	src := t[:]
+	src := t
 	dst := u[:]
 
 	for i, byteGroup := range byteGroups {


codec.go
```diff
@@ -125,7 +125,7 @@ func (u *UUID) decodeCanonical(t []byte) error {
 		return fmt.Errorf("uuid: incorrect UUID format %s", t)
 	}
 
-	src := t[:]
```
adamdecaf: Could this be an optimization? I recall a talk from today talking about "this being on the stack" now IIRC.

Can we benchmark before and after?

zerkms: Both are values on stack I presume

jadr2ddude: The t value is already a slice i think. The `[:]` should be a no-op.

zerkms: It looks like a no-op indeed https://go.godbolt.org/z/OuMHTy

theckman: What's the reason for even doing it? Might it make more sense just to put `src` in the function declaration and omit the line altogether? It's not a breaking API change.

acln0: @theckman Unfortunately, that would mean we lose consistency with the rest of the `decodeX` methods in that file, which all take a `t []byte` parameter.

FWIW, I have no opinion on this. I think the proposed change is fine as-is.

PR #57, opened 2018-10-06
Enhance defaultHWAddrFunc() and tests to hit 100% coverage

theckman:
This updates `defaultHWAddrFunc` to use a package-level variable for calling
`net.Interfaces()`, so that we can mock it in unit tests.

Doing so, and writing the related test, allows us to hit 100% coverage.


generator_test.go
```diff
@@ -376,6 +378,103 @@ func testNewV5DifferentNamespaces(t *testing.T) {
...
+	}{
+		{
+			n: "error",
+			fn: func() ([]net.Interface, error) {
+				return nil, errors.New("controlled failure")
+			},
+			e: "controlled failure",
+		},
+		{
+			n: "no_valid_hwaddr_returned",
```
acln0: Other subtests in this package are named in `PascalCase`, such as `FaultyRand`, `StaleEpoch`, etc. Let's keep this consistent and not use `snake_case`.

generator_test.go
```diff
@@ -376,6 +378,103 @@ func testNewV5DifferentNamespaces(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.n, func(t *testing.T) {
+			// set the netInterfaces variable (function) for the test
+			// and then set it back to default in the deferred function
+			netInterfaces = tt.fn
+			defer func() {
+				netInterfaces = net.Interfaces
+			}()
+
+			var hw net.HardwareAddr
```
acln0: As far as I can tell, these next four lines could be simplified to `hw, err := defaultHWAddrFunc()`, unless there is a stylistic reason you would like to keep them this way.

generator_test.go
```diff
@@ -376,6 +378,103 @@ func testNewV5DifferentNamespaces(t *testing.T) {
 	}
 }
 
+func Test_defaultHWAddrFunc(t *testing.T) {
```
acln0: I don't like this underscore very much, because no other top level tests have one. Indeed, `TestdefaultHWAttrFunc` does not look nice, but I wouldn't mind `TestDefaultHWAddrFunc`, even though the token after `Test` does not match the function being tested exactly.

I don't feel strongly about this at all, so we can leave it as-is if you'd like.

generator_test.go
```diff
@@ -376,6 +378,103 @@ func testNewV5DifferentNamespaces(t *testing.T) {
...
+						HardwareAddr: net.HardwareAddr{5, 6, 7, 8},
+					},
+				}
+
+				return s, nil
+			},
+			e: "uuid: no HW address found",
+		},
+		{
+			n: "valid_hwaddr_returned",
```
acln0: `snake_case` vs. `PascalCase` once again.

generator.go
```diff
@@ -284,9 +284,11 @@ func newFromHash(h hash.Hash, ns UUID, name string) UUID {
 	return u
 }
 
+var netInterfaces = net.Interfaces
```
acln0: At the call site for `netInterfaces`, we use `ifaces, err := netInterfaces()` and rely on type inference. I would prefer adding an explicit type here, in the style of these hooks: https://golang.org/pkg/internal/poll/#pkg-variables, to better communicate the signature of what is being hooked, so the reader doesn't have to go through yet another indirection.

Perhaps `var netInterfaces func() ([]net.Interface, error) = net.Interfaces`.

generator_test.go
theckman: I am asserting that we didn't accidentally change types.

generator_test.go
theckman: Ugh. I didn't notice that in the original PR and would have asked for a change. That's against my personal style. Would anyone be against me going through and retrofitting the tests?

generator.go
theckman: I don't really like how it clutters up the code. It would be easier for my editor to give me the function header definition over trying to read that line.

I was trying to hit 100% without making the code too messy.

generator_test.go
theckman: Although that seems extremely heavy handed. My personal style ended up being `snake_case` to  improve readability of test outputs.

acln0: My personal style is `PascalCase` and it was me who converted most of the tests to their current form, back when we got rid of gocheck, so I guess you can blame me! But if you feel strongly about `snake_case`, I can live with that too, so long as we have consistency. If you want to change all the subtests to `snake_case`, go ahead.

generator_test.go
acln0: Okay, that's fine by me.

generator.go
acln0: I want to push back a little here. I like having the type inline, be it a little bit more cluttered as it may. Editors might help alleviate the extra indirection via highlighting, but I'd like to avoid it altogether.

generator_test.go
adamdecaf: I'm indifferent FWIW.

theckman:
Sorry for the delay, October was busy for me. Let me do some homework on this one.

generator.go
theckman: Finally getting back around to working on this PR, and this change results in a `golint` failure.

```
generator.go:287:19: should omit type func() ([]net.Interface, error) from declaration of var netInterfaces; it will be inferred from the right-hand side (golint)
var netInterfaces func() ([]net.Interface, error) = net.Interfaces
                  ^
```

generator_test.go
tariq1890: Agree with @acln0 , Let's have `TestDefaultHWAddrFunc`

cameracker, line 381:
```suggestion
func TestDefaultHWAddrFunc(t *testing.T) {
```

cameracker (changes requested):
Left a code suggestion for the one finding that I think has been blocking the PR 😆

cameracker (approved):

PR #58, opened 2018-10-26
Change (*UUID) Scan to accept type UUID argument

nanjj:
Type UUID can not used as primary key, which is not NULL in table. Type
NullUUID can be used as reference to the primary key in another table,
which may be NULL. If we use jinzhu/gorm to implement the usage
scenario, jinzhu/gorm convert UUID to NullUUID via NullUUID Scan
call. Without the fix below error may be found: uuid: cannot convert
uuid.UUID to UUID.

This fix changes (*UUID) scan to accept UUID argument.

### DIFF
--- a/sql.go
+++ b/sql.go
@@ -38,6 +38,10 @@ func (u UUID) Value() (driver.Value, error) {
 // a longer byte slice or a string will be handled by UnmarshalText.
 func (u *UUID) Scan(src interface{}) error {
 	switch src := src.(type) {
+	case UUID: // support gorm convert from UUID to NullUUID
+		*u = src
+		return nil
+
 	case []byte:
 		if len(src) == Size {
 			return u.UnmarshalBinary(src)
--- a/sql_test.go
+++ b/sql_test.go
@@ -118,6 +118,7 @@ func TestNullUUID(t *testing.T) {
 	t.Run("Scan", func(t *testing.T) {
 		t.Run("Nil", testNullUUIDScanNil)
 		t.Run("Valid", testNullUUIDScanValid)
+		t.Run("UUID", testNullUUIDScanUUID)
 	})
 
 	t.Run("MarshalJSON", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


acln0:
Hello, and thank you for sending this PR! I'm a little bit confused about it though, so please help me understand. The https://golang.org/pkg/database/sql/#Scanner documentation lists the possible types of `src`, and `UUID` is not one of those types. Are you hitting a case in which `src` is of type `uuid.UUID`? If so, how? Is an ORM involved?

nanjj:
Yes, I am using jinzhu/gorm, see below structs:
```
type A struct {
UUID uuid.UUID
}
type B struct {
...
AID uuid.NullUUID
}
```
AID is NullUUID type (in order to make it nullabe in db), which refer A.UUID( which is UUID since it's primary, which can not be null in db).

sql.go
```diff
@@ -74,6 +74,11 @@ func (u *NullUUID) Scan(src interface{}) error {
 		return nil
 	}
 
+	if id, ok := src.(UUID); ok {
```
acln0: On the code path at line 84, we delegate to `UnmarshalText` or `UnmarshalBinary`, which validate the UUID. Here, we do no such thing, and we take it for granted that if `src` is of type `UUID`, then it represents a valid UUID.

In the context of your ORM and its usage, is there ever a world in which this `src` is not a valid UUID?

acln0: In any case, I think we should leave a comment here, which describes why we are doing this. It should mention that certain ORMs use `sql.Scanner` to convert Go values to other Go values.

nanjj:
@acln0 jinzhu/gorm yes, `src` can be `[]byte` or `string` in jinzhu/gorm, and can be `UUID` type.

sql.go
```diff
@@ -38,6 +38,10 @@ func (u UUID) Value() (driver.Value, error) {
 // a longer byte slice or a string will be handled by UnmarshalText.
 func (u *UUID) Scan(src interface{}) error {
 	switch src := src.(type) {
+	case UUID: // support gorm convert from UUID to NullUUID
+		u = &src
```
acln0: This is almost certainly not right. It changes the pointer `u`'s value, but not the UUID it points to. `*u = src`.

sql_test.go
```diff
@@ -193,6 +194,21 @@ func testNullUUIDScanValid(t *testing.T) {
 	}
 }
 
+func testNullUUIDScanUUID(t *testing.T) {
+	s := UUID{}
```
acln0: Let's not use the zero value here. It could lead to confusion. Please use `codecTestUUID` instead.

sql.go
nanjj: done

sql_test.go
nanjj: done

PR #60, opened 2018-11-16
Update README.md

komuw:
add the correct canonical way of importing go modules that are greater than or equal to version 2

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,8 @@ package main
 import (
 	"log"
 
-	"github.com/gofrs/uuid/v3"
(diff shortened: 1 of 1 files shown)


jadr2ddude:
Maybe:
```
"github.com/gofrs/uuid"
// if using modules: "github.com/gofrs/uuid/v3"
```

acln0:
Let's hold off on this until #61 is resolved. It seems tricky.

theckman:
We're about to merge a change to remove module support, which deprecates this PR. Thank you for the contribution and I'm sorry we weren't able to use it. We'll definitely resurrect something similar in the future depending on how modules shake out.

PR #62, opened 2018-11-20
unify import paths under /v4

### LINKED ISSUE #61: Unstable Import Paths
With the introduction of the `go.mod` file and declaration of `github.com/gofrs/uuid/v3` there is no longer a single import path that works correctly for both modules and non-modules users.

This has become an issue with gobuffalo/pop and other Buffalo projects.

Below you find a selection of `go get` statements that represent the different import paths and what happens when trying to use them in different environments.

### Go >1.9.7 (or `GO111MODULE=off`)

```bash
$ go get github.com/gofrs/uuid/v3

package github.com/gofrs/uuid/v3: cannot find package "github.com/gofrs/uuid/v3" in any of:
    /usr/local/go/src/github.com/gofrs/uuid/v3 (from $GOROOT)
    /go/src/github.com/gofrs/uuid/v3 (from $GOPATH)
```

### Go 1.11 (`GO111MODULE=on`)

```bash
$ go get github.com/gofrs/uuid
```

```go.mod
module github.com/gobuffalo/uuidmods

require github.com/gofrs/uuid v3.1.0+incompatible // indirect
```

```bash
$ go get github.com/gofrs/uuid/v3
```

```go.mod
module github.com/gobuffalo/uuidmods

require (
	github.com/gofrs/uuid/v3 v3.1.2 // indirect
)
```

acln0:
Create a new /v4 directory, containing a copy of the existing code,
allowing the package to move forward and be consumed by all users
under a unified import path: github.com/gofrs/uuid/v4.

Remove go.mod file from the top level directory and add it to the v4
directory. Update v3 to v4 in go.mod.

Update documentation to reflect the change: Recommend v4.0.0+ as the
go-to version for new users and existing modules users. Recommend
v2.1.0 under github.com/gofrs/uuid as the go-to version for users
looking for a replacement for satori/go.uuid at v1.2.0. Document the
two different import paths and their behavior.

Fixes #61.


acln0:
You're welcome to try to talk the actual users out of it in #61, then.

jadr2ddude:
Could we use a symlink?

zerkms:
@jadr2ddude on windows git would not create a symlink

jadr2ddude:
Ok. Ugh.

myitcv:
@theckman 

> Please, no.

Can you be more specific about what it is you are objecting to here?

@acln0 has, to my mind, taken the right step of distilling a concrete straw-man PR out of the advice from @bcmills and @thepudds in #61. Constructive comments on where that straw-man deviates from the advice would probably be more useful. Else, if it's the advice you disagree with, I'd suggest we continue the discussion in #61.

theckman:
@myitcv I believe in "A little copying is better than a little dependency", but I'm adamantly against versioned subdirectories and the subsequent code copying. I want to make use of the repository's metadata for version information and not directory structure.

myitcv:
@theckman 

> but I'm adamantly against versioned subdirectories and the subsequent code copying

Did you see https://github.com/gofrs/uuid/issues/61#issuecomment-440283403 which confirms:

> Note that /v4 doesn't have to be a copy. It can forward functions and types back to the non-/v4 path, or you can move (not copy) the code to /v4 and rewrite the non-/v4 path to forward to /v4. Either would reduce the maintenance overhead, since bug-fixes in one path would automatically apply to the other as well. (That's what the goforward tool is for.)

acln0:
Abandoning in favor of #66.

PR #63, opened 2018-11-22
Use the latest Go releases

cristaloleg:
Should we also drop support for older(pre 1.10) versions? 
They're not patched nor supported anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
 language: go
 sudo: false
 go:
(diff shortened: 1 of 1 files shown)


acln0:
Thanks for the contribution. What we do for CI moving forward depends on the decision we make for #61, so I think it's wise to wait a little while with this.

cristaloleg:
Roger that

Issue #65, opened 2018-12-03
de-virtualize the /v3 path, remove go.mod from root


acln0:
As discussed in the #gofrs Slack channel, perhaps we finally have a resolution for #61.

The plan is to de-virtualize the `/v3`, remove `go.mod` from the project root, and add forwarding declarations and type aliases in the `/v3` subdirectory, followed by cutting v3.2.0.

If anyone is willing to do this work and submit it for review, that would be great. Please ping me when you do. I had previously sent https://github.com/gofrs/uuid/commit/c6f696b8aab1385872bdb55a01e019d46574e9b8 which is the other way around. There, the forwarding package is the root package, while `/v3` contains the actual code. Perhaps it would be good to base the work on that commit, but we want to do it the other way around. The root package should be the canonical one, and we bite the bullet for a while to support existing modules users using the `/v3` subdirectory.

What happens with a potential v4.0.0 in the future remains to be seen.

itsjamie:
I'll do this in the next four hours.

PR #66, opened 2018-12-04
Creates a virtual package at v3 subdirectory.

itsjamie:
Step 1 in removing go modules opt-in support without breaking any
builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ env:
 before_install:
   - go get golang.org/x/tools/cmd/cover
 script:
-  - go test ./... -race -coverprofile=coverage.txt -covermode=atomic
+  - go test . -race -coverprofile=coverage.txt -covermode=atomic
 after_success:
   - bash <(curl -s https://codecov.io/bash)
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: go
 sudo: false
 go:
-  - 1.7
-  - 1.8
   - 1.9
   - "1.10"
   - 1.11
--- a/v3/forward.go
+++ b/v3/forward.go
@@ -19,7 +19,7 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-package v3
+package uuid
 
 import (
 	"github.com/gofrs/uuid"
--- a/v3/forward.go
+++ b/v3/forward.go
@@ -1,0 +1,183 @@
+// Copyright (c) 2018 Andrei Tudor Călin <mail@acln.ro>
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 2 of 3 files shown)


itsjamie:
Copied forward.go implementation to package.

I renamed the package identifier to match the folder name... but I'm realizing this breaks the whole idea, so changing to be uuid and deal with the duplicate name.

itsjamie:
```
package main

import (
	"github.com/davecgh/go-spew/spew"
	"github.com/gofrs/uuid/v3"
)

func main() {
	u, _ := uuid.NewV4()
	spew.Dump(u)
}
```

works with `GO111MODULE=off go run`

itsjamie:
See #65 .

dylan-bourque:
I tested this change against a trivial "client" (import the `uuid` package and print the result of `uuid.NewV4()` to stdout) in Go 1.8.7, 1.9.7, 1.10.5, and 1.11.1.  For each version, I ran `go build` and `go run main.go` for each combination of: with/without the `/v3` in the import path, with/without `go.mod`.  I also installed Dep and ran `dep init` to check backwards compatibility w/ vendoring.  Additionally, I created a "client" outside of `$GOPATH` under Go 1.11 with a `go.mod`.

All cases were good except importing `github.com/gofrs/uuid/v3` under Go 1.8, but this is expected because type aliases are Go 1.9+.

acln0:
Thanks @itsjamie I think this set of patches is ready now. Usually, I would squash such a set down to a single commit, to end up with a nicer git history, but it's up to you if you want to do it or not.

After this is merged, someone should cut and publish v3.2.0. Thanks.

thepudds:
@itsjamie @acln0 sorry if this is off base (including I might have lost the thread here a bit on this general topic given there have been more than a few comments)... but I had thought the `v3/go.mod` here was going to end up with a `require` directive for `github.com/gofrs/uuid` to help enforce user a user importing `github.com/gofrs/uuid/v3` ends up with the latest `github.com/gofrs/uuid`?

See for example this comment https://github.com/gofrs/uuid/issues/61#issuecomment-441751134 from Bryan Mills:

> The forwarding itself should work fine, but if you're doing active development in a part of the tree that lacks a go.mod file, you'll need to periodically bump the requirements in the /v3 module to pick up the changes. (Probably every time you push a new tag to the +incompatible path you'll want to add that same tag — and the corresponding require directive — to the v3/go.mod file.)

(CC @bcmills or @myitcv for any additional comment)

.travis.yml
```diff
@@ -16,7 +14,7 @@ env:
 before_install:
   - go get golang.org/x/tools/cmd/cover
 script:
-  - go test ./... -race -coverprofile=coverage.txt -covermode=atomic
+  - go test . -race -coverprofile=coverage.txt -covermode=atomic
```
myitcv: Just a thought, but is it worth expanding the tests to include an explicit test of the behaviour you're expecting a user to see, with respect to import paths? That suite of tests might also include your expectations on GOPATH vs module mode in Go 1.11.

thepudds:
I think part of the earlier intent of this change was also to encourage people to use `github.com/gofrs/uuid` over `github.com/gofrs/uuid/v3`?

For example, slightly older snippet from @acln0:

> We ask them to remove `/v3` from their import paths _eventually_, and switch to `gofrs/uuid` instead. `gofrs/uuid` remains `+incompatible` with modules, and all future development happens in the root of the directory. After a while, we phase `/v3` away ...

I don't personally know if that is indeed is still the intent, but if it is the intent, would it make sense to add a quick note to the README to explain something like that?

For example, maybe something relatively brief, such as:

> **Note**: the recommended import path for all non-module and module-based consumers is `import "github.com/gofrs/uuid"`.  (For legacy reasons, `github.com/gofrs/uuid/v3` is also currently supported, but any consumer using that `/v3` import path is  encouraged to migrate to importing `github.com/gofrs/uuid`).

Or maybe that is no longer the intent?

acln0:
Thanks @thepudds and @myitcv for the pertinent comments. Maybe this isn't quite ready yet. Holding off until we resolve them. I'll find some time to work on this very soon.

dylan-bourque:
The "break" came from adding `go.mod` then making further changes on the
same major version. That led to inconsistent imports when using `
github.com/gofrs/uuid` vs `github.com/gofrs/uuid/v3`. IMO, the big miss on
the part of the Go team was not loudly calling out that you need to do a
major version bump when opting into modules if your package is already v2+.

And I'm not really a fan of creating subfolders for each version but it's
the only way for module-unaware importers to get semver behavior. I could
get behind doing it for v3 for compatibility during this transition, but
not really going forward.

Re: logrus, if we're taking it over we'll more than likely want to do this
same dance there.

itsjamie:
After conversation in the #gofrs Slack, I'm going to close this pull-request in favour of removal of the `go.mod` file.

itsjamie:
I was confused.

PR #71, opened 2019-02-15
Remove support for DCE Security UUID (V2) generation

theckman:
This change removes our support for generating V2 UUIDs from the package, since
it seems that the current implementation is not up to spec in relation to the
DCE 1.1 Specification[1][2].

In our implementation the data written in to the V1, V3-V5 UUID uses Big Endian
byte order as specified in RFC-4122[3]. However, when looking at the DCE
specification they make it read like the data should be written in not Little
Endian but a format where you completely reverse the bites of the value from
most significant first to least significant first:

> Set the time_low field equal to the least significant 32-bits (bits numbered
> 0 to 31 inclusive) of the time stamp in the same order of significance. If a
> DCE Security version UUID is being created, then replace the time_low field
> with the local user security attribute as defined by the DCE: Security
> Services specification.

There are more things later confirming the order:

> Set the time_mid field equal to the bits numbered 32 to 47 inclusive of the
> time stamp in the same order of significance.

This timestamp handling instructions seem dangerous because the specification is
telling us to overwrite the lest-significant bits of the UUID timestamp value
with a deterministic value, meaning the remaining bits of the timestamp will now
change much less frequently. Because the rest of the UUID consists of distinct /
deterministic values, this will make the UUID much less unique.

This also means that for the DCE Security UUIDs we would need completely
different generating and parsing code than V1 and V3-V5 UUIDs. Considering that
it seems like there would be quite a bit of work to support them correctly.

To confirm that my understanding of the spec is correct, I tried to look at
preexisting implementations to validate my understanding. Unfortunately, I've
not had luck finding a UUID library that implements V2 UUIDs. That points to
there being little to no value in continuing to try and support them, as there
is not widespread availability of complementing implementations.

This change will require a major version bump.

[1] http://pubs.opengroup.org/onlinepubs/9629399/apdxa.htm#tagcjh_20_02_05
[2] http://pubs.opengroup.org/onlinepubs/9668899/chap5.htm#tagcjh_08_02_01_01
[3] https://tools.ietf.org/html/rfc4122#section-4.1.2

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -19,11 +19,19 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-// Package uuid provides implementations of the Universally Unique Identifier (UUID), as specified in RFC-4122 and DCE 1.1.
+// Package uuid provides implementations of the Universally Unique Identifier
+// (UUID), as specified in RFC-4122,
 //
 // RFC-4122[1] provides the specification for versions 1, 3, 4, and 5.
 //
-// DCE 1.1[2] provides the specification for version 2.
+// DCE 1.1[2] provides the specification for version 2, but version 2 support
+// was removed from this package in v4 due to some concerns with the
+// specification itself. Reading the spec, it seems that it would result in
+// generating UUIDs that aren't very unique. In having read the spec it seemed
+// that our implementation did not meet the spec. It also seems to be at-odds
+// with RFC 4122, meaning we would need quite a bit of special code to support
+// it. Lastly, there were no Version 2 implementations that we could find to
+// ensure we were understanding the specification correctly.
 //
 // [1] https://tools.ietf.org/html/rfc4122
 // [2] http://pubs.opengroup.org/onlinepubs/9696989899/chap5.htm#tagcjh_08_02_01_01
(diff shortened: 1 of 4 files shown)


adamdecaf:
There look to be some callers, but it sounds like they're broken anyway? 

https://github.com/search?q=language%3Ago+%22uuid.NewV2%22&type=Code

uuid.go
```diff
@@ -46,7 +46,7 @@ type UUID [Size]byte
 const (
 	_  byte = iota
 	V1      // Version 1 (date-time and MAC address)
-	V2      // Version 2 (date-time and MAC address, DCE security version)
+	_       // Version 2 (date-time and MAC address, DCE security version) [removed]
```
jadr2ddude: I would leave the constant there. There is no reason to stop users from being able to identify that this is a v2 UUID.

```suggestion
	V2      // Version 2 (date-time and MAC address, DCE security version) [not supported]
```

uuid.go
```diff
@@ -68,7 +68,7 @@ const (
 )
 
 // Timestamp is the count of 100-nanosecond intervals since 00:00:00.00,
-// 15 October 1582 within a V1 UUID. This type has no meaning for V2-V5
+// 15 October 1582 within a V1 UUID. This type has no meaning for V3-V5
```
jadr2ddude: I would simplify this a bit:

```suggestion
// 15 October 1582 within a V1 UUID. This type has no meaning for other
```

uuid.go
theckman: What makes you believe removing this stops users from being able to identify a UUIDv2? I think it makes more sense for them to now define a constant in their project instead of rely on ours for it, since we no longer provide V2 support.

theckman: To clarify, the type here is a `byte`. It should be trivial for someone to define their own constant and use it. It would also avoid people using our package just for the `V2` constant.

jadr2ddude: This isn't that important, so it is fine to remove.

PR #72, opened 2019-02-15
Implement fmt.Formatter

dylan-bourque:
I had a use case where I needed to get the string representation of a UUID with only the hex digits.  Implementing `fmt.Formatter` seems to be the prescribed way to provide custom string formatting in Go, so I did so.  Additionally, `UUID.String()` returns the RFC-compliant string so I did not touch that code.

The standard `%s`, `%q` and `%v` format runes have the same behavior as before.  I added `%h` and `%H` for outputting only the hex digits without the dash (-) separators.  `%h` uses 'a'-'f' and `%H` uses 'A-F'


uuid.go
```diff
@@ -156,6 +158,47 @@ func (u UUID) String() string {
...
 
+// Format implements fmt.Formatter for UUID values.  It accepts 'h' or 'H' as a format to output only
+// the hex digits of the UUID.  Any other format specifier returns the canonical RFC-4122 representation.
+func (u UUID) Format(f fmt.State, c rune) {
+	//
+	// TODO: dylan.bourque - 2019-02-15
+	// . should we implement %v? if so, what is "a Go-syntax representation" of a UUID when using %#v?
+	//
+	switch c {
+	case 'h', 'H':
```
acln0: If we are going to do this here, I think the verbs should be `%x` and `%X`, rather than `%h` and `%H`, since the former verbs already mean "print in hexadecimal" in other contexts.

The circumstances are a little unfortunate. Observe this program: https://play.golang.org/p/uRc8V3pnuEa

Unfortunately, out of the box, it seems like `%x` takes the output from `UUID.String` and encodes it in hexadecimal. I'd venture to say (with relative confidence) that nobody relies on this output today, so we could change `%x` to do The Right Thing instead of what it does out now.

That being said, even without adding new code to package uuid, one possible fix to get `%x` to do that is to use `.Bytes()`, as the program shows.

uuid.go
```diff
@@ -156,6 +158,47 @@ func (u UUID) String() string {
...
+		s := hex.EncodeToString(u.Bytes())
+		if c == 'H' {
+			s = strings.Map(toCapitalHexDigits, s)
+		}
+		_, _ = io.WriteString(f, s)
+	case 'q':
```
acln0: I don't think this is necessary. I think we're getting this behavior out of the box: https://play.golang.org/p/GcHtLq7aZxl

The `fmt` package documentation also talks about this: 

> If the format (which is implicitly %v for Println etc.) is valid for a string (%s %q %v %x %X), the following two rules apply:
> 
> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).
> 
> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

uuid.go
```diff
@@ -156,6 +158,47 @@ func (u UUID) String() string {
...
+		_, _ = io.WriteString(f, `"`+u.String()+`"`)
+	default:
+		_, _ = io.WriteString(f, u.String())
```
acln0: I'm not convinced about this approach.

When `UUID` does not implement `fmt.Formatter`, this program fails the `go vet` check, and produces obviously wrong output: https://play.golang.org/p/stts25e3_2a

However, when `UUID` _does_ implement `fmt.Formatter`, the program passes `go vet`, and produces more plausible looking (but wrong) output: https://play.golang.org/p/hNOR123ZjkU

The behavior in the second case seems dangerous to me. Using the wrong formatting verb is a programmer error and this implementation of `fmt.Formatter` hides it from `go vet`, and swallows it when it produces output. I don't know what a good solution to this problem might be, short of trying to re-create the original format string from the `fmt.State` and calling `fmt.Sprintf` with a `[16]byte` value instead of a `UUID`, to avoid recursion.

For reference, `pkg/errors` implements `fmt.Formatter` on some of its types, and when it doesn't understand a verb, it simply produces no output, e.g. https://github.com/pkg/errors/blob/master/stack.go#L64-L84. I'm not a huge fan of this approach either, but it is at least a data point we have, if we're going to implement `fmt.Formatter` at all.

acln0:
To clarify: I'm not sure we should be doing this, but if we decide to, then I think it's not ready yet, and we need some good answers, especially to the third comment I left above.

cratermoon:
What's wrong with the caller using strings.Replace()? https://play.golang.org/p/fPJoSEQjZ1t

dylan-bourque:
@cratermoon Technically, nothing. Your example generates the desired result. But it loses context (because returning the hex digits of a UUID is not what `strings.Replace()` does). It also results in several unnecessary allocations while building the string with the dashes only to immediately remove them.

@acln0 I'll go through your suggestions tomorrow. Thanks for the thorough feedback.

uuid.go
dylan-bourque: I'm fine with %x and %X rather than %h and %H.

You are correct that the default behavior of %x is to take the string RFC-format string returned by `.String()` and hex-encode the byte values of those characters, which I would say is definitively wrong and should be fixed.

My thoughts on using `fmt.Sprintf("%x", u.Bytes())` is the same as for `strings.Replace(u.String(), "-", "", -1)`, namely that there's no logical abstraction.  While correct, either of these amounts to forcing consumers to dig into the internals of the instance.

uuid.go
dylan-bourque: I tested this case and you must account for the %q verb in your `.Format()` implementation.  If you don't handle it directly, the result is whatever you do in the default case.  See here -> https://play.golang.org/p/FNja_rxReEx

uuid.go
dylan-bourque: There doesn't seem to be any way to indicate that a formatting verb is invalid/not supported, so it feels like "do nothing" is the only option.  I'm okay with that path for all of the formatting verbs that don't apply: %b, %c, %d, %o, %U, %e, %E, %f, %F, %g, %G.

dylan-bourque:
As an alternative to implementing `fmt.Formatter` and all of the vagueness that seems to be involved with doing so, we could also implement `.Format()` directly a la `time.Time`, either as a package-level free function or as a method on the `UUID` type.

uuid.go
acln0: Ah. I was not clear enough, sorry. What I meant to say is that if we're not implementing `fmt.Formatter`, we're getting this behavior for `%q` anyway, due to `UUID` implementing `fmt.Stringer` and the rule I quoted above.

But if we _do_ implement `fmt.Formatter`, we should honor `%q` explicitly indeed.

uuid.go
acln0: To clarify, "do nothing" means "emit no output"?

uuid.go
acln0: It seems like we have consensus on %x and %X, then. I think that's good enough for this particular point, so I'm going to mark it as resolved.

uuid.go
dylan-bourque: Correct, "do nothing" would be to return an empty string for all of those verbs.

inliquid:
```go
hex.EncodeToString(u.Bytes())
```

dylan-bourque:
@inliquid

dylan-bourque:
@inliquid 
> The idea was to provide a higher level logical abstraction.

uuid.go
acln0: Okay. I guess that's reasonable, and there is a precedent for it too. But my concern with respect to the fact that this conceals programmer errors such as using the wrong flags remains.

The `fmt.Formatter` implementation is opaque to static analysis tools like `go vet`. We're gaining convenience and a nice API for printing UUIDs as hexadecimal strings, but we are losing precision under the lens of static analysis tools. We should decide if this trade-off is worth making.

dylan-bourque:
I updated `.Format()` to use 'x' and 'X' instead of 'h' and 'H' for only the hex digits and updated the logic to return an empty string for unsupported format verbs.

dylan-bourque:
The current implementation of `UnmarshalText()` is already coded to allow for the "hex digits, no separators, lowercase letters" format this PR proposes to add with the `%x` formatting verb, so a discussion about limiting the API strictly to what's defined in the RFC would need to consider that code also.I don't know that we _should_ limit the functionality to only what's defined in the RFC.  Personally, I lend more weight to whether or not the functionality is logical and useful without breaking compliance with the underlying RFC specification.  I think this PR adheres to all three of those conditions.

As far as usefulness, I've had a need for encoding/decoding UUID values in this format on two separate occasions now.  I submitted the PR based on a guess that others may need/want this same behavior also, without punting to "You can do it yourself if you just ....".

I have no issue with the library implementing a superset of the RFC but I am far from the only one with an investment and/or opinion here.

uuid.go
```diff
@@ -156,6 +158,53 @@ func (u UUID) String() string {
...
+	switch c {
+	case 'x', 'X':
+		s := hex.EncodeToString(u.Bytes())
+		if c == 'X' {
+			s = strings.Map(toCapitalHexDigits, s)
+		}
+		_, _ = io.WriteString(f, s)
+	case 'v', 's':
+		// TODO: dylan.bourque - 2019-02-18
+		// . should we implement %#v? if so, what is "a Go-syntax representation" of a UUID?
```
theckman: I'd say let's do it. I wonder if we need to coerce the value in to a `[Size]byte` and then return the `%#v` of that.

uuid.go
```diff
@@ -156,6 +158,53 @@ func (u UUID) String() string {
...
+// All other verbs not handled directly by the fmt package (like %p) return an empty string.
+func (u UUID) Format(f fmt.State, c rune) {
+	switch c {
+	case 'x', 'X':
+		s := hex.EncodeToString(u.Bytes())
+		if c == 'X' {
+			s = strings.Map(toCapitalHexDigits, s)
+		}
+		_, _ = io.WriteString(f, s)
+	case 'v', 's':
```
theckman: How would you feel about scope-creeping a bit here, and adding a `%S` which returns the RFC format in all capital letters? Would there be any objections for this?

theckman:
This looks good. Wondering if we can add two other features to this to make it :100:. 

I'm thinking I could hold the PR removing `V2` support, and get this in to a `3.x` release.

uuid.go
dylan-bourque: I'll give that a shot and see what it feels like. Sounds like the right path.

dylan-bourque:
I'll try to make these changes some time today

uuid.go
dylan-bourque: The `%v` and `%+v` verbs still return the default RFC 4122 string.  Which format would be preferable for `fmt.Sprintf("%#v", u)`?
```
[Size]byte(u):
    "[16]uint8{0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef, 0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef}"
u.Bytes():
    "[]byte{0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef, 0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef}"
```

FWIW, I'm partial to the latter.

acln0: `%#v` means Go syntax.

If `%#v` produced the string `"foo"` for type `Foo`, users should be able to type `var f Foo = foo` in their Go programs, and that code should compile.

theckman: We'd want the `[16]uint8` format, because that's the real type:

```
const Size = 16

type UUID [Size]byte
```

dylan-bourque: Under that rule, the output should be `uuid.UUID([]byte{.....})`, right?  But then it's "broken" if someone aliases the package name on import. :man_shrugging: #catch22

acln0: Step 1: https://play.golang.org/p/ocJNNBjEiI2

Step 2: https://play.golang.org/p/s_dA-yquJZd

Since Step 2 compiles, that seems like the only real option to me.

dylan-bourque: Implemented `%#v` as `[16]byte{....}`.

Also changed the default case to return `"%!verb(type=value)"` as recommended by the documentation in the `fmt` package (See "Format Errors" under https://golang.org/pkg/fmt/#hdr-Printing), i.e. `fmt.Sprintf("%d", u)` will return `%!d(uuid.UUID=....)`.

acln0: > Also changed the default case to return `"%!verb(type=value)"` as recommended by the documentation in the fmt package

Very nice!

uuid_test.go
```diff
@@ -114,6 +115,43 @@ func testUUIDSetVariant(t *testing.T) {
...
+		{u: val, f: "%F", want: "%!F(uuid.UUID=12345678-90ab-cdef-1234-567890abcdef)"},
+		{u: val, f: "%g", want: "%!g(uuid.UUID=12345678-90ab-cdef-1234-567890abcdef)"},
+		{u: val, f: "%G", want: "%!G(uuid.UUID=12345678-90ab-cdef-1234-567890abcdef)"},
+		{u: val, f: "%o", want: "%!o(uuid.UUID=12345678-90ab-cdef-1234-567890abcdef)"},
+		{u: val, f: "%U", want: "%!U(uuid.UUID=12345678-90ab-cdef-1234-567890abcdef)"},
+	}
+	for _, tt := range tests {
+		got := fmt.Sprintf(tt.f, tt.u)
+		if tt.want != got {
+			t.Errorf("Format(\"%s\") got %s, want %s", tt.f, got, tt.want)
```
tariq1890:
```suggestion
			t.Errorf(`Format("%s") got %s, want %s`, tt.f, got, tt.want)
```

uuid.go
```diff
@@ -156,6 +158,65 @@ func (u UUID) String() string {
...
+// All other verbs not handled directly by the fmt package (like '%p') are unsupported and will return
+// "%!verb(uuid.UUID=value)" as recommended by the fmt package.
+func (u UUID) Format(f fmt.State, c rune) {
+	switch c {
+	case 'x', 'X':
+		s := hex.EncodeToString(u.Bytes())
+		if c == 'X' {
+			s = strings.Map(toCapitalHexDigits, s)
+		}
+		_, _ = io.WriteString(f, s)
```
tariq1890: Do we really need the dashes here? Why not do this ?
```suggestion
		io.WriteString(f, s)
```

dylan-bourque: This was intentional b/c I have my linter set to warn on ignored return values.  I prefer to explicitly call out that I don't care about the return values over leaving them silently discarded.

martin308:
@theckman any plans to cut a new release with this change in?

theckman:
@martin308 3.4.0 will be out within the next 30 minutes.

theckman:
@martin308 https://github.com/gofrs/uuid/releases/tag/v3.3.0

Sorry for the delay, a few of us could've sworn we released it. Thanks for the ping!

martin308:
Amazing, thank you! 🙏

PR #78, opened 2019-05-09
improve codec error strings

coder543:
Right now, the invalid format error strings are confusing for attempts to parse empty strings, since they just end in a `length: `. It really looks like it should be printing the invalid length.

Furthermore, some of the errors use colons, and some of them don't, so even for the exact same error, the strings are inconsistent.

I propose to add single quotes around any `%s` verbs to make error messages more readable, as well as printing the erroneous length value and making colon usage consistent within `codec.go`

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -114,7 +114,7 @@ func (u *UUID) UnmarshalText(text []byte) error {
 	case 41, 45:
 		return u.decodeURN(text)
 	default:
-		return fmt.Errorf("uuid: incorrect UUID length: %s", text)
+		return fmt.Errorf("uuid: incorrect UUID length %d in string %q", len(text), text)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


codec.go
```diff
@@ -114,15 +114,15 @@ func (u *UUID) UnmarshalText(text []byte) error {
 	case 41, 45:
 		return u.decodeURN(text)
 	default:
-		return fmt.Errorf("uuid: incorrect UUID length: %s", text)
+		return fmt.Errorf("uuid: incorrect UUID length %d in string '%s'", len(text), text)
```
acln0: What if we used `%q` instead of `'%s'`?

acln0 (changes requested):
Sounds good to me in principle. One concern, though.

codec.go
coder543: I actually wasn't aware of the `%q` verb. That's neat! I switched the code to use `%q` instead of `'%s'`

acln0:
LGTM now! In the interest of clean git history, can you please squash these down into one commit, and then force-push?

coder543:
done!

acln0 (approved):
Thanks for the contribution! Waiting on someone else to review this too, and then we can merge.

adamdecaf (approved):

adamdecaf:
I'm assuming anyone who cares about this specific error is matching on `uuid: incorrect UUID length` (perhaps without the `uuid:` prefix).

niaow (approved):
LGTM

acln0:
Thanks again. Merging.

PR #79, opened 2019-06-22
Modify codec to avoid heap allocations in the happy path

ztstewart:
UUIDs are implemented as a fixed length array, giving great performance.

However, while profiling, I've found a few cases in the parsing code that can lead to heap allocations:

1. Passing values to fmt.Errorf, which accepts interface{}, which typically requires that parameters be boxed
2. Converting []byte to string or vice versa

We can avoid these by doing the following:

1. Copy values passed to fmt.* and friends, which will speed up the happy path
2. Using some unsafe trickery for []byte <--> string conversion

This PR implements both of these strategies. As a result, various parsing methods are faster.

Old benchmark results:

```
 % go test -benchmem -bench ./...
goos: linux
goarch: amd64
pkg: github.com/gofrs/uuid
BenchmarkFromString/canonical-8                 10000000               123 ns/op              48 B/op          1 allocs/op
BenchmarkFromString/urn-8                       10000000               133 ns/op              48 B/op          1 allocs/op
BenchmarkFromString/braced-8                    10000000               130 ns/op              48 B/op          1 allocs/op
BenchmarkGenerator/NewV1-8                       1000000              1092 ns/op               0 B/op          0 allocs/op
BenchmarkGenerator/NewV2-8                       1000000              1095 ns/op               0 B/op          0 allocs/op
BenchmarkGenerator/NewV3-8                       5000000               273 ns/op             144 B/op          4 allocs/op
BenchmarkGenerator/NewV4-8                       1000000              1488 ns/op              16 B/op          1 allocs/op
BenchmarkGenerator/NewV5-8                       5000000               317 ns/op             176 B/op          4 allocs/op
PASS
ok      github.com/gofrs/uuid   11.589s
```

New benchmark results:

```
 % go test -benchmem -bench ./...
goos: linux
goarch: amd64
pkg: github.com/gofrs/uuid
BenchmarkFromString/canonical-8                 20000000                87.9 ns/op             0 B/op          0 allocs/op
BenchmarkFromString/urn-8                       20000000                97.6 ns/op             0 B/op          0 allocs/op
BenchmarkFromString/braced-8                    20000000                92.2 ns/op             0 B/op          0 allocs/op
BenchmarkGenerator/NewV1-8                       1000000              1110 ns/op               0 B/op          0 allocs/op
BenchmarkGenerator/NewV2-8                       1000000              1100 ns/op               0 B/op          0 allocs/op
BenchmarkGenerator/NewV3-8                       5000000               275 ns/op             144 B/op          4 allocs/op
BenchmarkGenerator/NewV4-8                       1000000              1485 ns/op              16 B/op          1 allocs/op
BenchmarkGenerator/NewV5-8                       5000000               322 ns/op             176 B/op          4 allocs/op
PASS
ok      github.com/gofrs/uuid   13.213s
````

Diff:

```
% benchcmp old.txt new.txt
benchmark                           old ns/op     new ns/op     delta
BenchmarkFromString/canonical-8     128           88.6          -30.78%
BenchmarkFromString/urn-8           134           101           -24.63%
BenchmarkFromString/braced-8        131           94.9          -27.56%
BenchmarkGenerator/NewV1-8          1072          1096          +2.24%
BenchmarkGenerator/NewV2-8          1078          1111          +3.06%
BenchmarkGenerator/NewV3-8          271           274           +1.11%
BenchmarkGenerator/NewV4-8          1471          1506          +2.38%
BenchmarkGenerator/NewV5-8          318           317           -0.31%

benchmark                           old allocs     new allocs     delta
BenchmarkFromString/canonical-8     1              0              -100.00%
BenchmarkFromString/urn-8           1              0              -100.00%
BenchmarkFromString/braced-8        1              0              -100.00%
BenchmarkGenerator/NewV1-8          0              0              +0.00%
BenchmarkGenerator/NewV2-8          0              0              +0.00%
BenchmarkGenerator/NewV3-8          4              4              +0.00%
BenchmarkGenerator/NewV4-8          1              1              +0.00%
BenchmarkGenerator/NewV5-8          4              4              +0.00%

benchmark                           old bytes     new bytes     delta
BenchmarkFromString/canonical-8     48            0             -100.00%
BenchmarkFromString/urn-8           48            0             -100.00%
BenchmarkFromString/braced-8        48            0             -100.00%
BenchmarkGenerator/NewV1-8          0             0             +0.00%
BenchmarkGenerator/NewV2-8          0             0             +0.00%
BenchmarkGenerator/NewV3-8          144           144           +0.00%
BenchmarkGenerator/NewV4-8          16            16            +0.00%
BenchmarkGenerator/NewV5-8          176           176           +0.00%
```

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -25,8 +25,6 @@ import (
 	"bytes"
 	"encoding/hex"
 	"fmt"
-
-	"github.com/gofrs/uuid/internal/unsafe"
 )
 
 // FromBytes returns a UUID generated from the raw byte slice input.
@@ -51,7 +49,7 @@ func FromBytesOrNil(input []byte) UUID {
 // Input is expected in a form accepted by UnmarshalText.
 func FromString(input string) (UUID, error) {
 	u := UUID{}
-	err := u.UnmarshalText(unsafe.StringToByteSlice(input))
+	err := u.UnmarshalText(stringToByteSlice(input))
 	return u, err
 }
 
@@ -68,7 +66,7 @@ func FromStringOrNil(input string) UUID {
 // MarshalText implements the encoding.TextMarshaler interface.
 // The encoding is the same as returned by the String() method.
 func (u UUID) MarshalText() ([]byte, error) {
-	return unsafe.StringToByteSlice(u.String()), nil
+	return stringToByteSlice(u.String()), nil
 }
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
--- a/codec.go
+++ b/codec.go
@@ -25,6 +25,8 @@ import (
 	"bytes"
 	"encoding/hex"
 	"fmt"
+
+	"github.com/gofrs/uuid/internal/unsafe"
 )
 
 // FromBytes returns a UUID generated from the raw byte slice input.
@@ -49,7 +51,7 @@ func FromBytesOrNil(input []byte) UUID {
 // Input is expected in a form accepted by UnmarshalText.
 func FromString(input string) (UUID, error) {
 	u := UUID{}
-	err := u.UnmarshalText(stringToByteSlice(input))
+	err := u.UnmarshalText(unsafe.StringToByteSlice(input))
 	return u, err
 }
 
@@ -66,7 +68,7 @@ func FromStringOrNil(input string) UUID {
 // MarshalText implements the encoding.TextMarshaler interface.
 // The encoding is the same as returned by the String() method.
 func (u UUID) MarshalText() ([]byte, error) {
-	return stringToByteSlice(u.String()), nil
+	return unsafe.StringToByteSlice(u.String()), nil
 }
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
--- a/codec.go
+++ b/codec.go
@@ -24,7 +24,9 @@ package uuid
 import (
 	"bytes"
 	"encoding/hex"
+	"errors"
 	"fmt"
+	"strings"
 )
 
 // FromBytes returns a UUID generated from the raw byte slice input.
@@ -49,7 +51,7 @@ func FromBytesOrNil(input []byte) UUID {
 // Input is expected in a form accepted by UnmarshalText.
 func FromString(input string) (UUID, error) {
 	u := UUID{}
-	err := u.UnmarshalText([]byte(input))
+	err := u.UnmarshalText(stringToByteSlice(input))
 	return u, err
 }
 
@@ -66,7 +68,7 @@ func FromStringOrNil(input string) UUID {
 // MarshalText implements the encoding.TextMarshaler interface.
 // The encoding is the same as returned by the String() method.
 func (u UUID) MarshalText() ([]byte, error) {
-	return []byte(u.String()), nil
+	return stringToByteSlice(u.String()), nil
 }
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
@@ -114,7 +116,15 @@ func (u *UUID) UnmarshalText(text []byte) error {
 	case 41, 45:
 		return u.decodeURN(text)
 	default:
-		return fmt.Errorf("uuid: incorrect UUID length %d in string %q", len(text), text)
+		// Copying text will prevent the Go compiler from allocating text on
+		// the heap if text would otherwise be on the stack. This can
+		// significantly speed up parsing and using UUIDs that are
+		// well-formed.
+		// On the other hand, if we have an error, returning the error string
+		// may be slightly slower.
+		tc := make([]byte, len(text))
+		copy(tc, text)
+		return fmt.Errorf("uuid: incorrect UUID length %d in string %q", len(tc), tc)
 	}
 }
 
@@ -122,7 +132,11 @@ func (u *UUID) UnmarshalText(text []byte) error {
 // "6ba7b810-9dad-11d1-80b4-00c04fd430c8".
 func (u *UUID) decodeCanonical(t []byte) error {
 	if t[8] != '-' || t[13] != '-' || t[18] != '-' || t[23] != '-' {
-		return fmt.Errorf("uuid: incorrect UUID format in string %q", t)
+		b := strings.Builder{}
+		b.Grow(64 + len(t))
+		b.WriteString("uuid: incorrect UUID format in string ")
+		b.Write(t)
+		return errors.New(b.String())
 	}
 
 	src := t
@@ -160,7 +174,11 @@ func (u *UUID) decodeBraced(t []byte) error {
 	l := len(t)
 
 	if t[0] != '{' || t[l-1] != '}' {
-		return fmt.Errorf("uuid: incorrect UUID format in string %q", t)
+		// Copying t will prevent the Go compiler from allocating t on
+		// the heap if t would otherwise be on the stack.
+		tc := make([]byte, len(t))
+		copy(tc, t)
+		return fmt.Errorf("uuid: incorrect UUID format in string %q", tc)
 	}
 
 	return u.decodePlain(t[1 : l-1])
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: slice_conversion.go [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"unsafe"
+)
+
+// byteSliceToString uses pointer manipulation tricks to take a slice of bytes
+// and convert it into a string. It should be used to avoid copying the slice.
+// It should also cause the compiler's escape analysis to keep the slice on
+// the stack if it would otherwise be stack allocated.
+func byteSliceToString(b []byte) string {
```
ztstewart: I am unsure if these tricks work on versions of Go below 1.12. If they do not, I believe it is possible to rename this file so it's Go 1.12 specific. Using the `reflect` package is another way of achieving something similar.

slice_conversion.go
```diff
@@ -0,0 +1,41 @@
+// Copyright (C) 2013-2018 by Maxim Bublis <b@codemonkey.ru>
```
ztstewart: This is a brand new file, with no ties to the old satori UUID package. What copyright notice is appropriate? I copied and pasted it from another file, but I don't think it's what I should be using.

ztstewart:
The CI build passes now; apologies for the spam. 

As noted, this change improves the performance of `FromString` and similar methods by a decent margin. More importantly, it means that UUIDs can live on the stack when parsed from a string or bytes, saving garbage collection time.

A few things I encountered that tripped me up:

- I had trivially changed `fmt.Errorf` to `errors.New` on a line that is uncovered in `defaultHWAddrFunc`. My attempt to modify the code to cover it made UUID generation slower, so I've reverted my original change.
- On Go 1.9 and lower, `golang.org/x/tools/cmd/cover` does not work if there are subpackages. I wanted to put the slice logic in `internal/unsafe` which would be cleaner, but creating that package fails the build. If we're OK with dropping 1.9 and lower from CI or have a workaround I can move it around again.

Thanks for taking a look,
Zach

ztstewart:
@acln0 @theckman, I believe you two reviewed the last couple of PRs. I'd greatly appreciate it if you could take a look.

Thanks a bunch,
Zach

dylan-bourque:
My 2 cents ... resorting to `unsafe` (and introducing specifically called out UB and panics) for the sake of shaving a handful of nanoseconds off of generating formatted error messages is a net loss for a general purpose library.  The new code is also significantly more complex than before and, in my opinion, requires some intimate knowledge of the internals of Go's memory management to make sense of why anyone would even care.

I would argue that a system where those <20 nanoseconds make a difference should probably be implementing their own UUID type.

ztstewart:
HI @dylan-bourque, thanks for reading the PR, and I appreciate the feedback.

> My 2 cents ... resorting to unsafe (and introducing specifically called out UB and panics) for the sake of shaving a handful of nanoseconds off of generating formatted error messages is a net loss for a general purpose library.

I would argue the opposite regarding generality: without these optimizations, the library becomes less generally useful. Reducing allocations in the `NewFromString` path to 0 means that a UUID can be stack allocated, which means that at scale you can keep the garbage collector happy. Wrapping it into the library itself means everyone who uses it benefits without any changes in the calling code.

As for panics: Panics can crop up inside the library if Go changes how it lays out strings (unlikely, given how proposals related to that are tagged with `Go2`) or if the library starts modifying user input in `UnmarshalText`. Modifying the input  would be a bug, regardless of whether or not this PR is landed.

I'm happy to add tests for functions that return `[]byte` to guarantee that modifying the output does not cause panics.

> The new code is also significantly more complex than before and, in my opinion, requires some intimate knowledge of the internals of Go's memory management to make sense of why anyone would even care.

I think you raise a valid point: We want to keep the bar for contribution as low as possible. That means code that is easy to read and understand, and I agree that's what we should optimize for.

However, I think the changes here don't require much specialized knowledge outside of the `unsafe` code, though I've encountered variations on this trick in a few places (including the standard library). 

My hope is that the clarifying comments which I have added (I would love your feedback on them!) can help ensure that future contributors understand the motivation for the changes I've made in error logging code. I personally feel that  `make` and `copy` are not very complex, being built in functions, but if you have suggestions on how to make the changes less complex, I'd be very appreciative of your suggestions.

> I would argue that a system where those <20 nanoseconds make a difference should probably be implementing their own UUID type.

That's certainly an option, and one I'm willing to pursue if we cannot come to an agreement on how to move forward. However, the work required to significantly optimize this library is fairly small and helps the entire Go community. I feel it would be a loss to the wider community to introduce yet another UUID library, and I'd love to work with you to avoid that possibility.

Thanks again for you feedback,
Zach

dylan-bourque:
@ztstewart 

By "general purpose", I meant that I think the vast majority of this library's consumers (the general case) won't actually care about those few nanoseconds and/or whether or not those 16 bytes escape to the heap when a UUID parsing error occurs.
> I would argue the opposite regarding generality: without these optimizations, the library becomes less generally useful

I also have concerns that these micro-optimizations could well turn out to be architecture/OS/platform/Go version/etc. dependent.

For sure, though, this is my opinion - and you know what they say about those :-) - and I'm definitely not any sort of "owner" over the project.

codec.go
```diff
@@ -175,7 +191,11 @@ func (u *UUID) decodeURN(t []byte) error {
 	urnUUIDPrefix := t[:9]
 
 	if !bytes.Equal(urnUUIDPrefix, urnPrefix) {
-		return fmt.Errorf("uuid: incorrect UUID format in string %q", t)
+		// Copying t will prevent the Go compiler from allocating t on
+		// the heap if t would otherwise be on the stack.
+		tc := make([]byte, len(t))
```
lunemec, line 196: I believe this should be handled by compiler optimizations, rather than code changes (especially if the code is identical in all cases).

slice_conversion.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// byteSliceToString uses pointer manipulation tricks to take a slice of bytes
+// and convert it into a string. It should be used to avoid copying the slice.
+// It should also cause the compiler's escape analysis to keep the slice on
+// the stack if it would otherwise be stack allocated.
+//
+// NOTE: modifying the input slice will likely result in panics, undefined
+// behavior, or worse. Use it for read-only scenarios.
+func byteSliceToString(b []byte) string {
```
lunemec, line 35: This function is not used anywhere, why keep it here?

slice_conversion.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// stringToByteSlice uses pointer manipulation tricks to take a string and
+// return the underlying slice of bytes referenced by the string. This is
+// useful when wanting to avoid extra allocations.
+//
+// NOTE: modifying the returned slice will likely result in panics, undefined
+// behavior, or worse. Use this function only when a read-only view
+// of the string is required.
+func stringToByteSlice(s string) []byte {
```
lunemec, line 46: Will this be safe when used with stdlib's https://golang.org/pkg/bytes/#Reader for example? I can imagine usecases like that, or buffering of output, etc.

uuid.go
```diff
@@ -143,7 +147,7 @@ func (u UUID) Bytes() []byte {
 // String returns a canonical RFC-4122 string representation of the UUID:
 // xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx.
 func (u UUID) String() string {
-	buf := make([]byte, 36)
+	buf := [36]byte{}
```
lunemec, line 150: 👍

lunemec:
@dylan-bourque while I agree that code should be as simple as possible, this code is not overly complex. And in addition it is used only in marshaling of the UUID to output. I would be OK with this being merged if it can be proven safe.

dylan-bourque:
@lunemec I still think this change is, for lack of a better word, unnecessary.  It adds more lines of warning comments than actual code and is optimizing for stack vs. heap allocations done inside of the standard library (which are subject to change in later Go releases).  It also introduces the use of `unsafe` to what is a very simple library for the sake of those micro-optimizations.

I won't block this PR but I'm also not going to approve it.

lunemec:
@ztstewart I guess since you put effort into optimizing that this is something important for your use-case?

@dylan-bourque maybe adding this as compile tag (only on-demand) would satisfy you, and it would allow users who want extra fast UUID parsing to use the unsafe method?

dylan-bourque:
As I said, I won't block the change if others approve and are willing to accept it.

cameracker:
Thanks for the contribution here. For the most part, this review has sat idle for a while due to some strong objections about increase of code complexity and use of unsafe methods for unclear benefit. Going to close this. If anybody has any objections, we can continue to talk about the changes here, and the PR can be re-opened.

PR #80, opened 2019-07-15
fixed version 2 domain bits to be spec compliant

bionoren:
I noticed some instability with testNewV2DifferentAcrossCalls. Wikipedia says "Version-2 UUIDs are similar to version 1, except that the least significant 8 bits of the clock sequence are replaced by a "local domain" number", which you're trying to do on line 181, except that you've generated the v1 UUID using binary.BigEndian. Big endian says that the "The most significant byte (MSB) value ... is at the lowest address. The other bytes follow in decreasing order of significance" (wikipedia again).

Since the clockSeq is inserted at [8:], assigning the domain to [9] was replacing the *most* significant bits (the ones most likely to change), not the least significant as specified. This patch fixes that. Not that anyone actually uses v2 in practice... but it's nice to have passing tests :)

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -178,7 +178,7 @@ func (g *Gen) NewV2(domain byte) (UUID, error) {
 		binary.BigEndian.PutUint32(u[:], posixGID)
 	}
 
-	u[9] = domain
+	u[8] = domain
 
 	u.SetVersion(V2)
 	u.SetVariant(VariantRFC4122)


CameronAckermanSEL:
Hi @bionoren , thanks very much for the contribution. 

We also noticed the bugs with UUID V2 and after a long, careful analysis, we concluded that UUID V2 is not specified enough for us to correctly support it in this library, and we're planning on deprecating it in the next major version of the library. https://github.com/gofrs/uuid/pull/71

Is there a particular reason that you need to use UUID V2? I'm unsure if you're aware, but it doesn't have enough entropy to be a viable random id generation algorithm and is more for for repeatability and identification purposes.

What are your thoughts?

bionoren:
Oh, no, I have no interest in it. It was just causing flaky tests and I didn't want that to cause problems with our automated builds. If it's going away soon, great!

theckman:
> Not that anyone actually uses v2 in practice... but it's nice to have passing tests :)

I think this comment summarizes the state of things. Since this is now on others' radar, it's likely a good idea to go forth and merge the removal of V2 support completely and cut the next major release of the library to avoid people wasting their time on a contribution we don't plan to merge-in.

For context, I was hesitating on the merge because it's yet another major version bump and I was hopeful Modules would have their stuff sorted out before we had to do that. I also aborted the plans to fix the way it was generated, because I was worried that (depending how people use the library) it may be a breaking change.

PR #82, opened 2019-10-21
Adds Compare and UUID.Equals functions

jonfriesen:
Hello! I didn't see a section on contributions, I hope I'm not overstepping with this proposal.

This change adds two functions, a `Compare(UUID, UUID) int`, and UUID attached `UUID.Equal(UUID) bool`:

Compare takes two UUID's and returns -1 if the first UUID is lexically before the second, 0 if they are equal, and 1 if the first UUID is lexically after the second.

Equal takes a single UUID and compares to the attached returning true if they are equal and false if not. Equal makes use of Compare.

This implementation follows RFC 4122 where equality and order checks should be done by converting the UUID to a unsigned 128-bit integer and compared.

Tests included

Cheers!
Jon


jadr2ddude:
Equal used to be here, but was removed because you can just use the `==` operator. See #36.

jadr2ddude:
Also, comparing UUIDs should just be `bytes.Compare` I think.

jonfriesen:
@jadr2ddude thanks for responding! You make good points, my attempt to follow the RFC clouded more obvious solutions like `==` and `bytes.Compare` :stuck_out_tongue: 

feel free to close this PR

Issue #85, opened 2020-05-05
create go.mod and go.sum


lizthegrey:
This ensures that users can import it as a /v3 module without needing the +incompatible go mod syntax.

CameronAckermanSEL:
This will probably wait until a breaking change is made to the library, such as the removal of uuidv2. https://github.com/gofrs/uuid/issues/61

theckman:
Hi Liz,

Thank you for raising this. 

We currently have no interest in adopting Go Modules, as it would break people who use `dep`, `glide`, or any package manager that doesn't support SIV. I wish the Go Authors wouldn't have put us in this unfortunate position. Please see this comment for more context:

- https://github.com/gofrs/uuid/pull/74#issuecomment-468076053

PaluMacil:
I'm not sure if you have a specific threshold, @theckman for making a major release and considering mod to be the official choice. In the 2019 go survey, they said:

> Almost everyone in the Go ecosystem is now using modules, but some confusion around package management remains.

This meant 89% said they used go mod, and modules were first available in 1.11, so now that we are on 1.14. And finally, if someone has older ways of doing this, they have an easy path away from things like dep and Glide:

> go mod init automatically translates the required information from dep, glide, govendor, godep, and 5 other pre-existing dependency managers into a go.mod file that produces the equivalent build.

Personally I thought dep was awful and still feel frustrated when I think about the issues I tried to raise in that community, but I understand that some people didn't like all the choices made by go mod either, and there are a couple things I would have done differently. At this point, I'm far more concerned about seeing the community be able to see projects adopting the official way of doing things. When I see a library not using go.mod now, I always need to dig a little further to see if it's a dead project, and I appreciate your library a lot and don't want people to skip over it because they assume it isn't being maintained.

theckman:
Thank you for reaching out. If there's something that's preventing Modules users from working with our package we're happy to try and address that. I hope our responsiveness here helps quell any maintenance concerns. 🙂

I'm personally still preferring the conservative approach because of SIV. If that wasn't an issue here we'd have Modules support awhile back.

stellirin:
I'd also welcome adding modules here, perhaps the release of Go v1.16 could be a reasonable enough milestone, when modules will have been GA for three major releases?

As hinted at above, I actually did skip over this project in favour of `github.com/google/uuid`, simply because of module support and a few more recent commits in the Git history.

[...]

PR #88, opened 2020-12-29
Bump version to 420, enable Go modules support

Xe:
The number was chosen as a reference to a peaceful community revolving around the use of an entheogen. This will make this package compliant with Go modules in the future.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/gofrs/uuid/v420
+
+go 1.15


theckman:
While I think it would be amusing to fuck with people / troll by requiring them to have marijuana references all over their codebase, I'm going to objectively refer back to this comment from another issue:

https://github.com/gofrs/uuid/issues/85#issuecomment-626123986

PR #93, opened 2021-08-28
initial implementation of UUIDv6 and UUIDv7 based on RFC Draft Rev 1

theckman:
There is currently an RFC draft in progress[1] to add three new UUID formats,
versions 6, 7, and 8, that change the internal data format so that the generated
UUIDs are k-sortable. Version 8 is a special UUID format designed for future
implementations or extensions, and as a result isn't implemented here.

Because this implmentation is based on a draft RFC, they may be subject to
change in later revisions. As such, these new functions/methods include a
comment indicating they do not yet fall under the API compatibility guarantee of
the project, and changes to meet new revisions to the spec will be done in minor
releases.

[1] https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -26,6 +26,7 @@ import (
 	"crypto/rand"
 	"crypto/sha1"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"hash"
 	"io"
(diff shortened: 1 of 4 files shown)


theckman:
So I updated my commit but codecov did not update the comment to indicate I'd fixed the coverage decrease...

theckman:
There we go...

generator.go
```diff
@@ -182,7 +220,39 @@ func (g *Gen) NewV5(ns UUID, name string) UUID {
...
+	}
+
+	timeNow, clockSeq, err := g.getClockSequence()
+	if err != nil {
+		return Nil, err
+	}
+
+	binary.BigEndian.PutUint32(u[0:], uint32(timeNow>>28))   // set time_high
+	binary.BigEndian.PutUint16(u[4:], uint16(timeNow>>12))   // set time_mid
+	binary.BigEndian.PutUint16(u[6:], uint16(timeNow&0xfff)) // set time_low (minus four version bits)
```
cameracker, line 246: is it meaningful how we mask this? Also unless I'm mistaken we have 0xfff which covers more than 4 bits, wouldn't we just want 0x0?

generator.go
```diff
@@ -182,7 +220,39 @@ func (g *Gen) NewV5(ns UUID, name string) UUID {
...
+	binary.BigEndian.PutUint16(u[8:], clockSeq&0x3fff)       // set clk_seq_hi_res (minus two variant bits)
+
```
cameracker, line 248: Do we need to set u[9:] to clk_sec_low_res

generator.go
```diff
@@ -182,7 +220,39 @@ func (g *Gen) NewV5(ns UUID, name string) UUID {
...
+	u.SetVersion(V6)
+	u.SetVariant(VariantRFC4122)
```
cameracker, line 250: is this variant correct? It's looking for the value, but can't remember of go has iotas 0 indexed (which would make this 1) or if this is indeed the right value.

generator.go
```diff
@@ -210,6 +280,226 @@ func (g *Gen) getClockSequence() (uint64, uint16, error) {
...
+		return Nil, err
+	}
+
+	msec := (nano / 1000000) & 0xfff
+
+	d := (sec << 28)           // set unixts field
+	d |= (msec << 16)          // set msec field
+	d |= (uint64(seq) & 0xfff) // set seq field
+
+	binary.BigEndian.PutUint64(u[:], d)
```
cameracker, line 369: is grouping it this way going to result in the msec and seq bits being stored BigEndian? That may be a deviation from the spec, because only the nano bits are stored as big endian (at least according to my reading).

Note, this concern applies to all the v7 generators.

cameracker (changes requested):
Had a few questions, but this seems mostly correct!

generator.go
theckman: @cameracker `SetVariant` does the translation.

generator.go
theckman: The original spec (RFC 4122) specifies that values should be in network byte order (Big Endian):

> The fields are encoded as 16 octets, with the sizes and order of the fields defined above, and with each field encoded with the Most Significant Byte first (known as network byte order).  Note that the field names, particularly for multiplexed fields, follow historical practice.

This draft doesn't make any changes to that.

generator.go
theckman: This does that implicitly. `uint16` is represented as two bytes, so we use the `BigEndian` binary encoder to put the `uint16` starting at `u[8]`. So it writes both bytes out, covering the hi and lo bits.

cameracker (approved):

generator.go
cameracker: I've approved the PR, but maybe a comment to that effect would be nice

cameracker:
I guess one more question, do you think we'd benefit from a test that verifies these are sortable as we expect?

theckman:
@cameracker I could probably do that in one of the tests I currently have where I generate a bunch of UUIDs to actually exercise the clock sequence generator. I could fail the test if the previous UUID isn't `<` the current one.

I'll add that in tonight.

johanbrandhorst:
Do we need to update the README for this or are we waiting for stabilization?

Issue #92, opened 2021-05-24
consider adding IsZero() property


kevinburke1:
Right now to tell if a UUID is entirely empty, you need to do this fairly clunky operation:

```go
if bytes.Equal(myUUID[:], uuid.Nil[:]) {
    ...
}
```

It might be nice to add `IsZero()` - matching the API for `time.Time` - that will handle this for you.

theckman:
@kevinburkemeter since they are an array type you can just compare them with `==`. Any reason you can't so that here?

cameracker:
Confirmed that @theckman 's suggestion works https://play.golang.org/p/0ZFu7v97XNm

With that in mind, @kevinburkemeter do you think we still need this method? Seems easy enough for someone to do in user land and one less method for us to maintain in the API surface area,

izolate:
@theckman You can't always count on having access to the `uuid` package, for example within Go templates, it would be nice to do this:

```go
{{if .SomeUUID.IsZero}}
Some UUID is nil
{{end}}
```

+1 for the suggestion to add an `IsZero` method.

cameracker:
Thanks for the recommendation @kevinburke1 , this has been contributed by @jordan-chalupka in #95, and is out for release in `v4.2.0`.

PR #95, opened 2021-11-23
adding IsNil method to uuid

jordan-chalupka:
Adding an IsZero method to uuid.

#92

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -151,8 +151,8 @@ var (
 	NamespaceX500 = Must(FromString("6ba7b814-9dad-11d1-80b4-00c04fd430c8"))
 )
 
-// IsZero returns if the UUID is equal to the nil UUID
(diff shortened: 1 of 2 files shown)


uuid.go
```diff
@@ -151,6 +151,11 @@ var (
 	NamespaceX500 = Must(FromString("6ba7b814-9dad-11d1-80b4-00c04fd430c8"))
 )
 
+// IsZero returns if the UUID is equal to the nil UUID
+func (u UUID) IsZero() bool {
```
cameracker, line 155: How about `IsNil`?

jordan-chalupka: This makes sense, updated the method name to IsNil.

Thanks!

cameracker (approved):

cameracker:
Thanks very much for this contribution! Hopefully this is a nice shortcut for users.

PR #96, opened 2021-12-04
improve performance and reduce allocations of UUID methods

charlievieth:
This commit improves the performance and reduces the number of allocs of
most UUID methods and adds a new UUID.Parse() method for parsing string
encoded UUIDs.

Parsing string encoded UUIDs is now 2x faster and no longer allocates.
The NullUUID MarshalJSON() and UnmarshalJSON() methods have also been
improved and no longer call out to json.Unmarshal. The UUID.Format
method has been improved for common cases.

Benchmark results:
```
goos: linux
goarch: amd64
pkg: github.com/gofrs/uuid
cpu: Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz

name                        old time/op    new time/op    delta
UnmarshalText/canonical-16    47.7ns ± 0%    22.3ns ± 0%   -53.26%  (p=0.000 n=8+10)
UnmarshalText/urn-16          47.8ns ± 1%    22.3ns ± 0%   -53.32%  (p=0.000 n=10+10)
UnmarshalText/braced-16       47.8ns ± 1%    22.3ns ± 0%   -53.34%  (p=0.000 n=9+9)
ParseV4-16                    86.5ns ±10%    22.6ns ± 0%   -73.85%  (p=0.000 n=9+8)
NullMarshalJSON/Valid-16       308ns ±21%      72ns ±12%   -76.71%  (p=0.000 n=10+10)
NullMarshalJSON/Invalid-16    41.8ns ± 2%     1.4ns ± 0%   -96.59%  (p=0.000 n=10+8)
Format/s-16                    151ns ± 3%     143ns ± 7%    -5.18%  (p=0.003 n=10+10)
Format/S-16                    305ns ± 2%     161ns ± 2%   -47.38%  (p=0.000 n=10+9)
Format/q-16                    217ns ±12%     144ns ± 6%   -33.52%  (p=0.000 n=10+10)
Format/x-16                    170ns ±13%     123ns ± 1%   -27.95%  (p=0.000 n=10+10)
Format/X-16                    324ns ±11%     148ns ± 1%   -54.15%  (p=0.000 n=10+10)
Format/v-16                    156ns ± 4%     142ns ± 5%    -8.68%  (p=0.000 n=10+10)
Format/+v-16                   155ns ± 3%     142ns ± 6%    -8.67%  (p=0.000 n=10+10)
Format/#v-16                   894ns ± 1%     847ns ± 1%    -5.22%  (p=0.000 n=10+9)
String-16                     70.1ns ±36%    70.4ns ±29%      ~     (p=0.971 n=10+10)
FromBytes-16                  1.81ns ± 0%    1.81ns ± 0%    -0.15%  (p=0.010 n=8+9)
FromString/canonical-16       94.3ns ±20%    23.3ns ± 1%   -75.25%  (p=0.000 n=10+10)
FromString/urn-16             93.7ns ±11%    23.8ns ± 1%   -74.66%  (p=0.000 n=10+8)
FromString/braced-16          87.1ns ± 5%    23.5ns ± 1%   -73.03%  (p=0.000 n=9+10)
MarshalBinary-16              0.20ns ± 3%    0.20ns ± 1%      ~     (p=0.922 n=10+9)
MarshalText-16                 115ns ±25%      22ns ± 1%   -80.66%  (p=0.000 n=10+10)

name                        old alloc/op   new alloc/op   delta
UnmarshalText/canonical-16     0.00B          0.00B           ~     (all equal)
UnmarshalText/urn-16           0.00B          0.00B           ~     (all equal)
UnmarshalText/braced-16        0.00B          0.00B           ~     (all equal)
ParseV4-16                     48.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
NullMarshalJSON/Valid-16        160B ± 0%       48B ± 0%   -70.00%  (p=0.000 n=10+10)
NullMarshalJSON/Invalid-16     8.00B ± 0%     0.00B       -100.00%  (p=0.000 n=10+10)
Format/s-16                    48.0B ± 0%     48.0B ± 0%      ~     (all equal)
Format/S-16                    96.0B ± 0%     48.0B ± 0%   -50.00%  (p=0.000 n=10+10)
Format/q-16                    96.0B ± 0%     48.0B ± 0%   -50.00%  (p=0.000 n=10+10)
Format/x-16                    64.0B ± 0%     32.0B ± 0%   -50.00%  (p=0.000 n=10+10)
Format/X-16                     112B ± 0%       32B ± 0%   -71.43%  (p=0.000 n=10+10)
Format/v-16                    48.0B ± 0%     48.0B ± 0%      ~     (all equal)
Format/+v-16                   48.0B ± 0%     48.0B ± 0%      ~     (all equal)
Format/#v-16                    128B ± 0%       16B ± 0%   -87.50%  (p=0.000 n=10+10)
String-16                      48.0B ± 0%     48.0B ± 0%      ~     (all equal)
FromBytes-16                   0.00B          0.00B           ~     (all equal)
FromString/canonical-16        48.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
FromString/urn-16              48.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
FromString/braced-16           48.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
MarshalBinary-16               0.00B          0.00B           ~     (all equal)
MarshalText-16                 96.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)

name                        old allocs/op  new allocs/op  delta
UnmarshalText/canonical-16      0.00           0.00           ~     (all equal)
UnmarshalText/urn-16            0.00           0.00           ~     (all equal)
UnmarshalText/braced-16         0.00           0.00           ~     (all equal)
ParseV4-16                      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
NullMarshalJSON/Valid-16        4.00 ± 0%      1.00 ± 0%   -75.00%  (p=0.000 n=10+10)
NullMarshalJSON/Invalid-16      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
Format/s-16                     1.00 ± 0%      1.00 ± 0%      ~     (all equal)
Format/S-16                     2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.000 n=10+10)
Format/q-16                     2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.000 n=10+10)
Format/x-16                     2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.000 n=10+10)
Format/X-16                     3.00 ± 0%      1.00 ± 0%   -66.67%  (p=0.000 n=10+10)
Format/v-16                     1.00 ± 0%      1.00 ± 0%      ~     (all equal)
Format/+v-16                    1.00 ± 0%      1.00 ± 0%      ~     (all equal)
Format/#v-16                    2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.000 n=10+10)
String-16                       1.00 ± 0%      1.00 ± 0%      ~     (all equal)
FromBytes-16                    0.00           0.00           ~     (all equal)
FromString/canonical-16         1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
FromString/urn-16               1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
FromString/braced-16            1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
MarshalBinary-16                0.00           0.00           ~     (all equal)
MarshalText-16                  2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -22,8 +22,7 @@
 package uuid
 
 import (
-	"bytes"
-	"encoding/hex"
+	"errors"
 	"fmt"
 )
 
@@ -45,11 +44,77 @@ func FromBytesOrNil(input []byte) UUID {
 	return uuid
 }
 
+var errInvalidFormat = errors.New("uuid: invalid UUID format")
+
+func fromHexChar(c byte) byte {
+	switch {
+	case '0' <= c && c <= '9':
+		return c - '0'
+	case 'a' <= c && c <= 'f':
+		return c - 'a' + 10
+	case 'A' <= c && c <= 'F':
+		return c - 'A' + 10
+	}
+	return 255
+}
+
+// Parse parses the UUID stored in the string text. Parsing and supported
+// formats are the same as UnmarshalText.
+func (u *UUID) Parse(s string) error {
+	switch len(s) {
+	case 32: // hash
+	case 36: // canonical
+	case 34, 38:
+		if s[0] != '{' || s[len(s)-1] != '}' {
+			return fmt.Errorf("uuid: incorrect UUID format in string %q", s)
+		}
+		s = s[1 : len(s)-1]
+	case 41, 45:
+		if s[:9] != "urn:uuid:" {
+			return fmt.Errorf("uuid: incorrect UUID format in string %q", s[:9])
+		}
+		s = s[9:]
+	default:
+		return fmt.Errorf("uuid: incorrect UUID length %d in string %q", len(s), s)
+	}
+	// canonical
+	if len(s) == 36 {
+		if s[8] != '-' || s[13] != '-' || s[18] != '-' || s[23] != '-' {
+			return fmt.Errorf("uuid: incorrect UUID format in string %q", s)
+		}
+		for i, x := range [16]byte{
+			0, 2, 4, 6,
+			9, 11,
+			14, 16,
+			19, 21,
+			24, 26, 28, 30, 32, 34,
+		} {
+			v1 := fromHexChar(s[x])
+			v2 := fromHexChar(s[x+1])
+			if v1|v2 == 255 {
+				return errInvalidFormat
+			}
+			u[i] = (v1 << 4) | v2
+		}
+		return nil
+	}
+	// hash like
+	for i := 0; i < 32; i += 2 {
+		v1 := fromHexChar(s[i])
+		v2 := fromHexChar(s[i+1])
+		if v1|v2 == 255 {
+			return errInvalidFormat
+		}
+		u[i/2] = (v1 << 4) | v2
+	}
+	return nil
+}
+
 // FromString returns a UUID parsed from the input string.
 // Input is expected in a form accepted by UnmarshalText.
-func FromString(input string) (UUID, error) {
-	u := UUID{}
-	err := u.UnmarshalText([]byte(input))
+func FromString(text string) (UUID, error) {
+	var u UUID
+	err := u.Parse(text)
 	return u, err
 }
 
@@ -66,7 +131,9 @@ func FromStringOrNil(input string) UUID {
 // MarshalText implements the encoding.TextMarshaler interface.
 // The encoding is the same as returned by the String() method.
 func (u UUID) MarshalText() ([]byte, error) {
-	return []byte(u.String()), nil
+	var buf [36]byte
+	encodeCanonical(buf[:], u)
+	return buf[:], nil
 }
 
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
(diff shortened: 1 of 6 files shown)


charlievieth:
@cameracker @theckman I see you've both recently contributed to and reviewed PRs for this repo. Are you the right people to ping for a review here?

theckman:
Yep. I _may_ have some time this coming weekend to go over this. Not sure about Cameron.

charlievieth:
Thanks the for quick response and no hurries on the review, I just wanted to make sure someone was aware of this PR.

charlievieth:
@theckman any chance you have time for a review and if not is there someone else I should ping? Thanks!

TheSecEng:
Bump, this is a great improvement

LeonanCarvalho (approved):

uuid.go
```diff
@@ -182,22 +174,33 @@ func (u UUID) Bytes() []byte {
 	return u[:]
 }
 
+// encodeHex encodes the canonical RFC-4122 form of UUID u into the
+// first 36 bytes dst.
+func encodeHex(dst []byte, u UUID) {
```
dylan-bourque, line 179: Maybe call this `encodeCanonical()` for accuracy?  When I read `encodeHex` I'm expecting the output to be a string of hex digits.

codec.go
```diff
@@ -45,11 +44,77 @@ func FromBytesOrNil(input []byte) UUID {
...
+		return c - 'a' + 10
+	case 'A' <= c && c <= 'F':
+		return c - 'A' + 10
+	}
+	return 255
+}
+
+// Parse parses the UUID stored in the string text. Parsing and supported
+// formats are the same as UnmarshalText.
+func (u *UUID) Parse(s string) error {
```
dylan-bourque, line 63: Why not make `Parse()` delegate to `UnmarshalText()` or vice versa?  The conversion from `string` to `[]byte` or `[]byte` to `string` is almost always optimized out by the compiler and this is non-trivial code to be duplicating.

codec_test.go
```diff
@@ -90,7 +91,7 @@ type fromStringTest struct {
 }
 
 // Run runs the FromString test in a subtest of t, named by fst.variant.
-func (fst fromStringTest) Run(t *testing.T) {
+func (fst fromStringTest) TestFromString(t *testing.T) {
```
dylan-bourque, line 94: Why was this method renamed?  The convention with these tests is that each one has a `Run()` method that does the needful.

It would be preferable to add a 2nd `t.Run()` than rename the method and add a separate `TestUnmarshalText()` method with a single `t.Run()`.

sql.go
```diff
@@ -83,27 +83,30 @@ func (u *NullUUID) Scan(src interface{}) error {
 	return u.UUID.Scan(src)
 }
 
+var nullJSON = []byte("null")
```
dylan-bourque, line 86: Why declare a package-level variable that's only referenced once?

sql.go
```diff
@@ -83,27 +83,30 @@ func (u *NullUUID) Scan(src interface{}) error {
...
 func (u *NullUUID) UnmarshalJSON(b []byte) error {
-	if bytes.Equal(b, []byte("null")) {
+	if string(b) == "null" {
 		u.UUID, u.Valid = Nil, false
 		return nil
 	}
-
-	if err := json.Unmarshal(b, &u.UUID); err != nil {
-		return err
+	if n := len(b); n >= 2 && b[0] == '"' {
```
dylan-bourque, line 106: Is this removing leading and trailing quotes?  If so, it's quite unintuitive to my eye.

`bytes.Trim(b, "\"")` is functionally equivalent and would be much less "clever" (as in wouldn't require an explanation)

sql_test.go
```diff
@@ -318,3 +319,47 @@ func testNullUUIDUnmarshalJSONMalformed(t *testing.T) {
...
+
+func BenchmarkNullMarshalJSON(b *testing.B) {
+	b.Run("Valid", func(b *testing.B) {
+		u, err := FromString("6ba7b810-9dad-11d1-80b4-00c04fd430c8")
+		if err != nil {
+			b.Fatal(err)
+		}
+		n := NullUUID{UUID: u, Valid: true}
+		for i := 0; i < b.N; i++ {
+			n.MarshalJSON()
```
dylan-bourque, line 331: Benchmarking gotcha: the compiler is allowed to optimize this call away since nothing is using the return values.

dylan-bourque (commented):
The functional changes seem reasonable and the benchmark improvements are definitely a win.  I had some non-functional feedback, though.

sql_test.go
charlievieth: I don't believe that is true. This benchmark takes 72ns so it's clearly calling `MarshalJSON` and the Go standard library rarely use the result [`strings/strings_test.go#L332-L350)`](https://github.com/golang/go/blob/9f0234214473dfb785a5ad84a8fc62a6a395cbc3/src/strings/strings_test.go#L332-L350).

sql.go
charlievieth: `nullJSON` is declared here so that `NullUUID.MarshalJSON()` does not allocated when it nil/invalid. This is common practice: [`encoding/json/decode.go#L602`](https://github.com/golang/go/blob/9f0234214473dfb785a5ad84a8fc62a6a395cbc3/src/encoding/json/decode.go#L602).

codec.go
charlievieth: > The conversion from string to []byte or []byte to string is almost always optimized out by the compiler and this is non-trivial code to be duplicating.

In some cases this is true and the runtime can stack allocate the string so only needs to do a memmove (see: [`slicebytetostring()`](https://github.com/golang/go/blob/9f0234214473dfb785a5ad84a8fc62a6a395cbc3/src/runtime/string.go#L81-L116)), but that is not true here since `([]byte)(s)` escapes to heap (you can check this by making the suggested change and running `go build -i -gcflags='-m'`).

sql.go
charlievieth: `bytes.Trim(b, "\"")` would make invalid  inputs that have either extraneous quotes or a quote on only one side valid since it would transform `""UUID""` to `UUID` and `"UUID` to `UUID`.

codec.go
charlievieth: Additionally, delegating to `UnmarshalText` could be DOS vector if this was used to parse untrusted input (a malicious actor could pass very large "UUID" causing the application to allocate large amounts of memory).

uuid.go
charlievieth: Done

codec_test.go
charlievieth: I don't have a strong opinion here, but the method was renamed because I added `TestUnmarshalText` to the `fromStringTest` type (and IMHO neither method should exist and the logic should actually live in the top-level test functions).

Since we can't have two `Run()` methods, we are testing two unique code paths, and duplicating this type and it's test cases (just to have a `Run()` method) seems like an undue maintenance burden. I thought it would be better to add the `TestUnmarshalText` method and rename `Run` to `TestFromString`.

> It would be preferable to add a 2nd t.Run() than rename the method and add a separate TestUnmarshalText() method with a single t.Run().

IMHO, invoking the `UnmarshalText` tests separately from the top-level `TestUnmarshalText` function makes for cleaner test output (e.g. `TestUnmarshalText/Valid/Canonical`. If we had two `Run` methods then the output (depending on how its implemented) would look something like:
```
TestFromString
TestFromString/Valid
TestFromString/Valid/Canonical
...
# UnmarshalText (test names are repeated)
TestFromString
TestFromString/Valid
TestFromString/Valid/Canonical
```

Or if we name the subtest "UnmarshalText" it'll look like:
```
TestFromString
TestFromString/Valid
TestFromString/Valid/Canonical
...
TestFromString/UnmarshalText
TestFromString/UnmarshalText/Valid
TestFromString/UnmarshalTextValid/Canonical
```

sql_test.go
charlievieth: Absolutely and this would actually be pretty performant if implemented with a generic function that `Parse` and `UnmarshalText` would delegate to (at the cost of dropping support for Go 1.17 and less).

codec_test.go
dylan-bourque: fwiw I don't have a strong opinion either.  My comment was mostly about sticking to the existing conventions, which pre-dated me.

sql_test.go
charlievieth: Maybe, but it would be a breaking change since it would invalidate a large number of existing benchmarks. So if introduced it would likely exempt benchmarks (and tests) from any such optimization (you have to do the same when writing a benchmark suite in C/C++ where the compilers are much more aggressive `#pragma GCC ...`).

dylan-bourque: I'm basing this on the _A note on compiler optimizations_ section of [this blog](https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go) from Dave Cheney.  The compiler choosing to eliminate that call is within the spec and, since the toolchain isn't covered by the Go1 compatibility guarantee, it wouldn't even be considered breaking.

codec.go
charlievieth: > My concern is that this is a fairly complex block of code that's identical to UnmarshalText() other than one processes string and the other []byte. It would be better in the long run to not have the duplication. 

I don't like the duplication either, but this is the only way to do this without allocating. The fact that we have 100% test coverage considerably alleviates these concerns IMO. In the long run this could be replaced by a generic function that takes both strings and byte slices, but that should only be done once Go 1.17 is no longer in wide use.

> What's the result if UnmarshalText() delegates to Parse() instead?

Same result and same issues around creating a copy of untrusted input.

> Also, since the logic is the same, any DOS vector in UnmarshalText() exists here as well. 

The DOS vector only exists if we delegate the call to `UnmarshalText/Parse` from `Parse/UnmarshalText` and have to allocate to create a copy of untrusted input.

sql_test.go
charlievieth: Yep, you're entirely correct the spec does not cover the toolchain and TIL in some cases the compiler will elide the function call unless the result it used (I checked this using a simple `square(x int) int` function, which being "pure" in the C sense is probably why the compiler decided it could be elided - though interestingly it does not elide the loop).

I'm happy to change this to assign to a dummy variable.

charlievieth:
@dylan-bourque if everything looks good, could I get another approval from @LeonanCarvalho or any other maintaner?

LeonanCarvalho:
> @dylan-bourque if everything looks good, could I get another approval from @LeonanCarvalho or any other maintaner?

I'm not a maintainer, just supporter , yours changes LGTM

LeonanCarvalho (approved):

codec_test.go
dylan-bourque: You've convinced me.  After another pass over the code I agree with the change, if for slightly different reasons (`fromStringTest` is the only type here with a `Run()` method and it's there to wrap executing the batch of tests.  we're doing 2 batches so 2 names does make sense).

dylan-bourque (approved):
LGTM.  Apologies for nit picking but I'm a big proponent of clear over clever and these kinds of optimizations have a strong tendency to lose clarity.

charlievieth:
@cameracker (or anyone else with write access) could I get this reviewed/merged?

cameracker:
@charlievieth  I'll give it a look over today with the plan to merge. Thanks so much <3

PR #98, opened 2022-02-25
Generator options

mlesar:
Hi, I had to generate UUIDv6 with custom timestamps, so I added generator options to be able to change the epoch function.
I'm creating this PR in case this is something that others might need or find useful.

Bye,
Matija

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -39,7 +39,7 @@ import (
 // UUID epoch (October 15, 1582) and Unix epoch (January 1, 1970).
 const epochStart = 122192928000000000
 
-type epochFunc func() time.Time
+type EpochFunc func() time.Time
 
 // HWAddrFunc is the function type used to provide hardware (MAC) addresses.
 type HWAddrFunc func() (net.HardwareAddr, error)
@@ -126,7 +126,7 @@ type Gen struct {
 
 	rand io.Reader
 
-	epochFunc     epochFunc
+	epochFunc     EpochFunc
 	hwAddrFunc    HWAddrFunc
 	lastTime      uint64
 	clockSequence uint16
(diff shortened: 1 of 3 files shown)


generator.go
```diff
@@ -137,13 +137,43 @@ type Gen struct {
...
 // interface check -- build will fail if *Gen doesn't satisfy Generator
 var _ Generator = (*Gen)(nil)
 
-// NewGen returns a new instance of Gen with some default values set. Most
-// people should use this.
-func NewGen() *Gen {
-	return NewGenWithHWAF(defaultHWAddrFunc)
+// NewGen returns a new instance of Gen with default values if no options are
+// passed.
+func NewGen(opts ...GenOption) *Gen {
```
theckman, line 147: As written this would be a breaking change for the module, and so I don't think we can accept this if it changes `NewGen()`. We should probably rename this to `NewGenWithConfig()`, and keep the `NewGen()` signature as-is.

Secondarily, I really dislike functional options and would prefer we do not adopt them for this use case. They are really hard to discover via documentation, where-as a configuration struct could be a lot more approachable / discoverable and still also support adding new options without it being a breaking change.

generator.go
```diff
@@ -39,7 +39,7 @@ import (
 // UUID epoch (October 15, 1582) and Unix epoch (January 1, 1970).
 const epochStart = 122192928000000000
 
-type epochFunc func() time.Time
+type EpochFunc func() time.Time
```
theckman, line 42: Now that this is being exported, could you please add a GoDoc-compatible comment that explains what this is:

```Go
// EpochFunc is ...
```

theckman (changes requested):

generator.go
mlesar: >As written this would be a breaking change for the module, 

Why it would be a breaking change if you can call it as before `uuid.NewGen()`?

>Secondarily, I really dislike functional options and would prefer we do not adopt them for this use case. They are really hard to discover via documentation, where-as a configuration struct could be a lot more approachable / discoverable and still also support adding new options without it being a breaking change.

Just to make sure I understand this, you would prefer to have it like this:
```go
type GenConfig struct {
	EpochFunc  EpochFunc
	HWAddrFunc HWAddrFunc
	RandReader io.Reader
}

func NewGen() *Gen {
	return NewGenWithConfig(GenConfig{})
}

func NewGenWithConfig(config GenConfig) *Gen {
	if config.EpochFunc == nil {
		config.EpochFunc = time.Now
	}
	if config.HWAddrFunc == nil {
		config.HWAddrFunc = defaultHWAddrFunc
	}
	if config.RandReader == nil {
		config.RandReader = rand.Reader
	}

	return &Gen{
		epochFunc:  config.EpochFunc,
		hwAddrFunc: config.HWAddrFunc,
		rand:       config.RandReader,
	}
}
```
?

I find it more specific when using options, as you just use the one that you want to adjust, with config it seems as if someone has to set all of the properties.

LeonanCarvalho:
I think this is a must since v6  supports custom timestamps, also very important for hi-precision implementations

cameracker:
Closing this PR as it appears to have been resumed by another contributor in #111

PR #99, opened 2022-06-05
update UUIDv7 implementation with RFC Draft Rev 03 spec

convto:
Currently, UUIDv6 and UUIDv7 implementations follow RFC draft rev 02 [1].
However, it has already been updated by RFC draft rev 03 [2] in March 2022.

Rev3 updates v6 and v7 and v8 specifications, but the v6 changes seem to
have no impact on implementation. Therefore, v6 and (unimplemented) v8 are
not affected, so only the UUIDv7 implementation will be updated.

[1] https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format-02
[2] https://datatracker.ietf.org/doc/html/draft-peabody-dispatch-new-uuid-format-03

In rev 03, complex sequences and variable precision timestamps have been removed from the specification.
The implementation is now simple and lock-free. It looks good.

This PR has breaking changes to the signature of the exported function `NewV7()` and `Gen.NewV7()` ,
but this does not seem to need consideration. I have referenced the following comment.
https://github.com/gofrs/uuid/blob/028e8409cdd0ed11a2b5bb3feb1ae2285ebb94fa/generator.go#L92-L97

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -26,7 +26,6 @@ import (
 	"crypto/rand"
 	"crypto/sha1"
 	"encoding/binary"
-	"errors"
 	"fmt"
 	"hash"
 	"io"
@@ -71,7 +70,7 @@ func NewV5(ns UUID, name string) UUID {
 // pseudorandom data. The timestamp in a V6 UUID is the same as V1, with the bit
 // order being adjusted to allow the UUID to be k-sortable.
 //
-// This is implemented based on revision 02 of the Peabody UUID draft, and may
+// This is implemented based on revision 03 of the Peabody UUID draft, and may
 // be subject to change pending further revisions. Until the final specification
 // revision is finished, changes required to implement updates to the spec will
 // not be considered a breaking change. They will happen as a minor version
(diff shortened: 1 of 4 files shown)


convto:
@theckman 
I saw you had recently implemented a Rev02 draft on https://github.com/gofrs/uuid/pull/93 PR!
I think you seem like a very good reviewer of this PR.

If you're the right member for the review, I'd appreciate your review if you have the time needed. :bow:

convto:
@cameracker
I see you have contributed a lot to this project.
If you have enough time, I would appreciate it if you could review this PR.

generator.go
```diff
@@ -280,244 +269,36 @@ func (g *Gen) getClockSequence() (uint64, uint16, error) {
...
-		panic(fmt.Sprintf("unknown precision value %d", p))
-	}
-
-	if err != nil {
+	if _, err := io.ReadFull(g.rand, u[6:]); err != nil {
 		return Nil, err
 	}
 
+	tn := g.epochFunc()
+	ms := uint64(tn.Unix())*1e3 + uint64(tn.Nanosecond())/1e6
```
cameracker, line 288: Can we simplify this by just doing UnixMilli()?

generator.go
```diff
@@ -280,244 +269,36 @@ func (g *Gen) getClockSequence() (uint64, uint16, error) {
...
+	u[0] = byte(ms >> 40)
+	u[1] = byte(ms >> 32)
+	u[2] = byte(ms >> 24)
+	u[3] = byte(ms >> 16)
+	u[4] = byte(ms >> 8)
+	u[5] = byte(ms)
```
cameracker, line 294: Do you think we benefit any from using the `binary.BigEndian.PutUint32` interface that we use for the v6? As written makes sense given the lack of a PutUint48 to cover the size of the time in milliseconds..

cameracker (approved):

cameracker (changes requested):
~Can we get the build to run?~

I see the passing build in the upstream repo, but would you mind rebasing to get the tests to run on two latest versions of go? The comments I left are questions, willing to merge with no changes but wanted to at least ask.

generator.go
convto: As you commented, I wrote this code because PutUint48 is not supported. v6 may have similar conditions as far as the specification is concerned.

However, there are some differences in the case of v6.
The v6 specification specifies 32bit time_high , 16bit time_mid, 16bit time_low_and_version separately, which may have the effect of conveying the layout more explicitly.

(v7 is specified in the specification as a single field, unix_ts_ms, so I have taken the approach described in the code...)

generator.go
convto: Well, it seems to be simpler to use UnixMilli() here!

The original code was receiving and processing sec and nanosec for flexibility, and I was writing the same way without being aware of it.
https://github.com/gofrs/uuid/blame/master/generator.go#L376-L381

Since we no longer need to worry about timestamp accuracy flexibility, it looks like we can use UnixMili()!

convto: Fixed to use UnixMilli() in this commit.
https://github.com/gofrs/uuid/pull/99/commits/5279295d1ce3f6556606edacdf9a2977ce004309

cameracker (approved):

convto:
@cameracker 
I don't have write access to the project, Could the maintainer members do the merging and releasing?

cameracker:
Thank you so much for the submission @convto  :) It's nice to have the WIP versions of the new uuid features in the library. Thanks for keeping them up to date 👍

PR #101, opened 2022-08-28
Add a New() method to wrap NewV4() with Must()

suber124:
The change is to add a New() method to just return a v4 UUID without error. It wraps the Must() method. Similar to this [New()](https://pkg.go.dev/github.com/google/uuid#New) API.

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -47,6 +47,11 @@ type HWAddrFunc func() (net.HardwareAddr, error)
 // DefaultGenerator is the default UUID Generator used by this package.
 var DefaultGenerator Generator = NewGen()
 
+// New returns a randomly generated UUID.
+func New() UUID {
+	return Must(DefaultGenerator.NewV4())
+}
+
 // NewV1 returns a UUID based on the current timestamp and MAC address.
 func NewV1() (UUID, error) {
 	return DefaultGenerator.NewV1()
--- a/generator_test.go
+++ b/generator_test.go
@@ -33,6 +33,7 @@ import (
 )
 
 func TestGenerator(t *testing.T) {
+	t.Run("New", testNew)
 	t.Run("NewV1", testNewV1)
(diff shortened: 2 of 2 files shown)


bryanbaek (approved):

peterstace:
This seems fairly asymmetrical to me. I don't think there's anything special about V4 UUIDs such they should get special treatment. If anything, this would make more sense to me named as `MustV4` (with corresponding methods added for other UUID versions as well).

suber124:
@bryanbaek 🥰

suber124:
> This seems fairly asymmetrical to me. I don't think there's anything special about V4 UUIDs such they should get special treatment. If anything, this would make more sense to me named as `MustV4` (with corresponding methods added for other UUID versions as well).

Hi @peterstace, thanks for the comment. 🥰 There is nothing special about the v4. Compared with the google/uuid [New()](https://pkg.go.dev/github.com/google/uuid#New) API, I just found gofrs/uuid is too verbose to use. I believe most of the time, people do not care which version of the UUID is used plus how we want to handle the error.

fred2167:
I also found generating methods are too verbose and I don't usually care which version I am using. A more compact signature is much appreciated

cameracker:
Thanks for the submission. Personally, I'm opposed to adding this to the API. The existing interface is still relatively compact, and it's fairly easy to add this in user land. Having the library provide a guided default to v4 is more opinionated than I think the library should be. 

I understand the sentiment that most people aren't going to care what uuid format they're using to get their entropy, but there are cases where it does matter, and it probably pays to be a deliberate with what format you're using. The IDs are different, and are optimized for different cases (especially the new v6-v7 ones).

cameracker:
Hi all, I'm going to close this PR. Thanks very much for the submission. If there's renewed interest, we can reopen it.

PR #104, opened 2022-10-26
stop using UnixMill() to make legacy versions compatible

convto:
The following PR makes use of UnixMill() to get milli-precision timestamps.
https://github.com/gofrs/uuid/pull/99

UnixMill() was added in go1.17, so it seems that the latest gofrs/uuid is not available from older versions of go.

According to [Go's release policy](https://go.dev/doc/devel/release#policy), the last two versions of go are supported.
Therefore, I don't think it is a critical issue that it is not available for go versions older than go1.16.
However, since we expect some users to use older versions of go, we thought it would be better to stop using UnixMill().

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -285,7 +285,7 @@ func (g *Gen) NewV7() (UUID, error) {
 	}
 
 	tn := g.epochFunc()
-	ms := uint64(tn.UnixMilli())
+	ms := uint64(tn.Unix())*1e3 + uint64(tn.Nanosecond())/1e6
 	u[0] = byte(ms >> 40)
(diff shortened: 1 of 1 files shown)


dylan-bourque (approved):

convto:
@cameracker 
In relation to the PR (#99) I sent out earlier, I have made some changes for compatibility from the old Go.

If you have time, I would appreciate your review.
If there are no problems with the PR, please release it.

cameracker (approved):

cameracker:
Thanks a ton for identifying this issue and submitting a fix!

Issue #107, opened 2022-12-22
`uuid.UUID.Value` panics with nil pointers


abdusco:
`uuid.UUID.Value` is not implemented with pointer receiver,
therefore it panics when passing nil values to an SQL query.

```
value method github.com/gofrs/uuid/v3.UUID.Value called using nil *UUID pointer
```

dylan-bourque:
This case is why [the `NullUUID` type](https://pkg.go.dev/github.com/gofrs/uuid#NullUUID) exists. It mirrors the behavior of the `Null*` types in the `database/sql` package.

cameracker:
Hi, I believe the API already supports this behavior as of some changes in its latest release. Please let us know if this is still a gap.

PR #108, opened 2022-12-22
fix: Implement sql.Valuer with pointer receiver

abdusco:
This allows interpreting `*uuid.UUID(nil)` values as `NULL` on SQL databases.

Closes: https://github.com/gofrs/uuid/issues/107

### DIFF
--- a/sql.go
+++ b/sql.go
@@ -29,7 +29,10 @@ import (
 )
 
 // Value implements the driver.Valuer interface.
-func (u UUID) Value() (driver.Value, error) {
+func (u *UUID) Value() (driver.Value, error) {
+	if u == nil {
+		return nil, nil
+	}
 	return u.String(), nil
 }
 
--- a/sql_test.go
+++ b/sql_test.go
@@ -22,13 +22,17 @@
 package uuid
 
 import (
+	"database/sql/driver"
 	"encoding/json"
 	"fmt"
 	"testing"
 )
 
 func TestSQL(t *testing.T) {
-	t.Run("Value", testSQLValue)
+	t.Run("Value", func(t *testing.T) {
+		t.Run("NotNil", testSQLValue)
+		t.Run("Nil", testSQLNilValue)
+	})
 	t.Run("Scan", func(t *testing.T) {
 		t.Run("Binary", testSQLScanBinary)
 		t.Run("String", testSQLScanString)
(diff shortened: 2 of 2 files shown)


charlievieth:
This is a breaking change since `uuid.UUID` values will no longer implement the `database/sql/driver.Valuer` interface.

abdusco:
> breaking change

How so? What needs to be changed by the users? 


```golang
package main

import (
	"database/sql/driver"
	"fmt"
)

type UUIDP struct {
}

func (u *UUIDP) Value() (driver.Value, error) {
	return "pointer", nil
}

type UUIDNP struct {
}

func (u UUIDNP) Value() (driver.Value, error) {
	return "non pointer", nil
}

func main() {
	// regular
	var u UUIDP
	fmt.Println(u.Value())
	var up *UUIDP
	// calling .Value on pointer implementation works ok.
	fmt.Println(up.Value()) // works

	if _, ok := ((any)(up)).(driver.Valuer); ok { // implements
		fmt.Printf("%T implements driver.Valuer\n", up)
	}

	// non-pointer
	var np UUIDNP
	fmt.Println(np.Value())
	var npp *UUIDNP
	if _, ok := ((any)(npp)).(driver.Valuer); ok { // implements
		fmt.Printf("%T implements driver.Valuer\n", npp)
	}
	// calling the implementation on a non-pointer method panics
	fmt.Println(npp.Value()) 
}

```
run it on playground: https://go.dev/play/p/6B2INaIbb0h

charlievieth:
> > breaking change
> 
> How so? What needs to be changed by the users?

This will break all code bases that expect a UUID value (not pointer) to implement `database/sql/driver.Valuer`. Add `var _ driver.Valuer = UUID{}` to your change and it will not compile. Basically, all code bases that pass UUID's as values will need to change their code to pass them as pointers (failures will sometimes occur at compile time, but will more often occur at runtime due to `interface{}`, which is delightfully bad). This is subtle and has to do with the how interfaces in Go work, which can be a bit weird.

dylan-bourque:
In addition to this being a breaking change as @charlievieth called out, `NullUUID` exists to cover nullable database columns following the established pattern of `database/sql.NullString`, etc.  This change would introduce ambiguity around how consumers are supposed to use the library.

cameracker:
Hi @abdusco  really appreciate that you identified this gap and took the time to submit a fix.

As others have stated, this is a breaking change to the library and as is I don't believe we can accept this donation. There's another PR that address a similar need (#113). I'll give that PR a couple of days before I merge it, would love to hear your thoughts over there. 

Thanks again!

PR #110, opened 2022-12-22
all: gofmt with Go 1.19

charlievieth:
Go 1.19 adds doc comments. This commit runs `gofmt` on all the code to update the comments.

https://tip.golang.org/doc/go1.19#go-doc

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -139,37 +139,36 @@ func (u UUID) MarshalText() ([]byte, error) {
 // UnmarshalText implements the encoding.TextUnmarshaler interface.
 // Following formats are supported:
 //
-//   "6ba7b810-9dad-11d1-80b4-00c04fd430c8",
-//   "{6ba7b810-9dad-11d1-80b4-00c04fd430c8}",
-//   "urn:uuid:6ba7b810-9dad-11d1-80b4-00c04fd430c8"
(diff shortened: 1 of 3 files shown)


dylan-bourque:
Changes seem 👍 on the surface.  What does the updated godoc actually look like, @charlievieth?  It should be fine since the indented blocks should render well in 1.18 or earlier but it's best to verify before moving forward.

charlievieth:
Verified that there is no change in presentation (I think pkg.go.dev might already be applying these changes - all that `gofmt` is now doing is trying to make the "semantic meaning" of comments clearer).

The changes here are just the result of `gofmt` (as of 1.19) adjusting the code blocks to match the expected format (there is also one change that adds the new build tag format but thats been around for a bit).

See the "Code Blocks" section of [doc/comment](https://tip.golang.org/doc/comment) for more information.

dylan-bourque (approved):

cameracker:
@charlievieth  thanks for taking the time to clean up the code here. Would you mind rerunning this against tip master? It looks like the major rework PR yielded some conflicts 👍

charlievieth:
> @charlievieth thanks for taking the time to clean up the code here. Would you mind rerunning this against tip master? It looks like the major rework PR yielded some conflicts 👍

Done and thanks for the quick review.

PR #111, opened 2022-12-27
Generator with options

LeonanCarvalho:
I rewrote the PR https://github.com/gofrs/uuid/pull/98 authored by @mlesar  based on the @theckman  comments and recommendations.

Help is wanted to make it happens, this is a very desirable feature especially to generate UUIDv6 with custom timestamps.

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -285,7 +285,7 @@ func (g *Gen) NewV7() (UUID, error) {
 	}
 
 	tn := g.epochFunc()
-	ms := uint64(tn.UnixMilli())
+	ms := uint64(tn.Unix())*1e3 + uint64(tn.Nanosecond())/1e6
 	u[0] = byte(ms >> 40)
 	u[1] = byte(ms >> 32)
 	u[2] = byte(ms >> 24)
--- a/generator.go
+++ b/generator.go
@@ -183,7 +183,9 @@ func NewGenWithOptions(opts ...GenOption) *Gen {
 	return gen
 }
 
-// WithHWAddrFunc is a GenOption that allows you to provide your own HWAddrFunc function
+// WithHWAddrFunc is a GenOption that allows you to provide your own HWAddrFunc
+// function.
+// When this option is nil, the defaultHWAddrFunc is used.
 func WithHWAddrFunc(hwaf HWAddrFunc) GenOption {
 	return func(gen *Gen) {
 		if hwaf == nil {
@@ -194,7 +196,9 @@ func WithHWAddrFunc(hwaf HWAddrFunc) GenOption {
 	}
 }
 
-// WithEpochFunc is a GenOption that allows you to provide your own EpochFunc function
+// WithEpochFunc is a GenOption that allows you to provide your own EpochFunc
+// function.
+// When this option is nil, time.Now is used.
 func WithEpochFunc(epochf EpochFunc) GenOption {
 	return func(gen *Gen) {
 		if epochf == nil {
@@ -205,7 +209,9 @@ func WithEpochFunc(epochf EpochFunc) GenOption {
 	}
 }
 
-// WithRandomReader is a GenOption that allows you to provide your own random reader
+// WithRandomReader is a GenOption that allows you to provide your own random
+// reader.
+// When this option is nil, the default rand.Reader is used.
 func WithRandomReader(reader io.Reader) GenOption {
 	return func(gen *Gen) {
 		if reader == nil {
--- a/generator.go
+++ b/generator.go
@@ -177,9 +177,7 @@ func NewGenWithOptions(opts ...GenOption) *Gen {
 	}
 
 	for _, opt := range opts {
-		if opt != nil {
-			opt(gen)
-		}
+		opt(gen)
 	}
 
 	return gen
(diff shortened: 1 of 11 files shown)


baldanca (approved):

generator.go
```diff
@@ -147,11 +151,57 @@ func NewGen() *Gen {
...
+		opt(gen)
+	}
+
+	return gen
+}
+
+// WithHWAddrFunc is a GenOption that allows you to provide your own HWAddrFunc function
+func WithHWAddrFunc(hwaf HWAddrFunc) GenOption {
+	return func(gen *Gen) {
+		gen.hwAddrFunc = hwaf
```
dylan-bourque, line 189: This needs a nil check.  `WithHWAddrFunc(nil)` will compile but lead to an immediate panic when the generator is used.

generator.go
```diff
@@ -147,11 +151,57 @@ func NewGen() *Gen {
...
+	}
+}
+
+// WithEpochFunc is a GenOption that allows you to provide your own EpochFunc function
+func WithEpochFunc(epochf EpochFunc) GenOption {
+	return func(gen *Gen) {
+		gen.epochFunc = epochf
```
dylan-bourque, line 196: Same here, need to check if `epochf` is `nil`.

generator.go
```diff
@@ -147,11 +151,57 @@ func NewGen() *Gen {
...
+	}
+}
+
+// WithRandomReader is a GenOption that allows you to provide your own random reader
+func WithRandomReader(reader io.Reader) GenOption {
+	return func(gen *Gen) {
+		gen.rand = reader
```
dylan-bourque, line 203: And here.

dylan-bourque (changes requested):

generator.go
LeonanCarvalho: We do not need to validate each option, instead, we can bypass `NewGenWithOptions` when the `nil` option has been passed. 

```go
func NewGenWithOptions(opts ...GenOption) *Gen {
	gen := &Gen{
		epochFunc:  time.Now,
		hwAddrFunc: defaultHWAddrFunc,
		rand:       rand.Reader,
	}

	for _, opt := range opts {
		if opt != nil {
			opt(gen)
		}
	}

	return gen
}
```

What do you think?

dylan-bourque: In the example I gave the option itself won't be `nil` but it would end up setting the `Gen.hwAddrFunc` field to `nil`.  The `if opt != nil` check will never be false unless someone explicitly passes `GenOption(nil)`.

To be 100% safe we should be checking the option itself in `NewGenWithOptions()` and checking for `nil` **values** in the option functions themselves.

```
func WithHWAddrFunc(hwaf HWAddrFunc) GenOption {
    return func(gen *Gen) {
        if hwaf != nil {
            gen.hwAddrFunc = hwaf
        }
    }
}
```
or
```
func WithHWAddrFunc(hwaf HWAddrFunc) GenOption {
    return func(gen *Gen) {
        if hwaf == nil {
            hwaf = defaultHWAddrFunc
        }
        gen.hwAddrFunc = hwaf
    }
}
```

LeonanCarvalho: Got it!

Changed

generator.go
```diff
@@ -147,11 +151,69 @@ func NewGen() *Gen {
...
 	}
+
+	for _, opt := range opts {
+		opt(gen)
+	}
+
+	return gen
+}
+
+// WithHWAddrFunc is a GenOption that allows you to provide your own HWAddrFunc function
```
dylan-bourque, line 186: I would document the behavior when/if `hwaf` is `nil`.  Something along the lines of "If hwaf is nil a default function is used that returns the host's MAC address."

Similar for the two other options.

LeonanCarvalho: Good point, fixed

dylan-bourque (approved):
LGTM

LeonanCarvalho:
@theckman @cameracker  It would be great to have your review too :)

generator_test.go
```diff
@@ -175,6 +190,33 @@ func testNewV1MissingNetworkFaultyRand(t *testing.T) {
...
+	g := NewGenWithOptions(
+		WithHWAddrFunc(func() (net.HardwareAddr, error) {
+			return []byte{}, fmt.Errorf("uuid: no hw address found")
+		}),
+		WithRandomReader(&faultyReader{
+			readToFail: 1,
+		}),
+	)
+
+	u, err := g.NewV1()
```
cameracker, line 203: any value in checking the error message?

LeonanCarvalho:
> @LeonanCarvalho thanks for the submission. What are your thoughts on the decrease of code coverage? Can we get back to 100%?

I think we could include tests for nil values since there is an if with default's fallback suggested here https://github.com/gofrs/uuid/pull/111#discussion_r1057721608


https://app.codecov.io/gh/gofrs/uuid/pull/111

generator_test.go
LeonanCarvalho: I can't imagine what else to check at this point. I just kept the same behaviour as the existing tests, I didn't want to modify anything pre-existing.

LeonanCarvalho:
Does something flaky on those tests? I ran it on my repo and it passed 
https://github.com/LeonanCarvalho/uuid/actions/runs/3788789705

LeonanCarvalho:
I found out that pkg x/tools cover is no longer active and marked as deprecated, that's why the test is falling https://pkg.go.dev/golang.org/x/tools/cmd/cover

LeonanCarvalho:
I opened another PR to solve this workflow issue https://github.com/gofrs/uuid/pull/115

Issue #97, opened 2021-12-09
Proposal: to release as a Go Module, archive repo and fork to github.com/gofrs/uuid.go@v0.4.2


theckman:
I floated this idea very roughly in the [\#gofrs](https://gophers.slack.com/archives/CBP4N9BEU) Slack channel, and it's to effectively rename this repository to avoid the Semantic Import Versioning (SIV) component of Go Modules

## Context

When this repository was forked from `github.com/sator/go.uuid`, we chose to keep the same versioning scheme as the original project to help the migration be easier for folks. We did not need to consider Go Modules at the time, and so it was easy to start the project at a major version higher than the original project and assume it was low cost to do a major version increase as necessary. 

Unfortunately, Go Modules has changed the situation, and so with hindsight it would have been better to not follow the versioning of the original project and to instead release as a `v0.x` version. 

## Proposal

We would create a new repository, `github.com/gofrs/uuid.go`, and push the `master` branch of this repository to that one. We would not push any of the tags from this repo there. We'd then create a new tag, `v0.4.2`, at the same commit† as `v4.2.0` of this repo. We would then update the README of this repo to point there, and archive the project. Any new development would happen there on v0.

† same commit + addition of `go.mod` file / any updates for the rename

cameracker:
I view this suggestion as a net negative. 

First, we would be asking our users to change their import paths to make this change, which is precisely what we are trying to avoid with adopting modules in the first place. 

Second, with a v0, we'd be communicating to users incorrectly that this is an unstable package when in reality this package is highly unlikely to experience any significant breaking changes to the interface in the near future. The only time we've had to make a breaking change was to remove the v2 features because they were inherently unstable. There is nothing on the horizon suggesting we'd do anything similar anytime soon.

Third, we'd lose context and history in this repo by switching over to a new name. We'd lose searchability and name recognition we've built over the last several years.

To me its preferable to make no changes and keep on going, or to bite the bullet and include modules in this current repository. There are a relatively small amount of people who are complaining about inclusion of modules. This seems not broken, I'm not inclined to fix it.

dylan-bourque:
[...]

PR #116, opened 2023-02-07
add gomod support

cameracker:
This PR adds gomod support as an alternate to #97, and implements #85.

In #97 , we discuss a proposal to fork the repo to stay perpetually at major version 0 so as to not force users to provide maintenance on their import paths. It's been a year since that proposal was opened and there's been no major steps have been made to move us to a different repo or incorporate modules. In the meantime, the go ecosystem has moved on to incorporate go.mod and it is increasingly making less sense for us to not comply with the standard package management for go.

Irrespective of the history this module has with go.mod (see notorious #61 and #65), the status quo has changed and consuming packages are no longer in flux with glide, dep, and gomod in a way that we need to support all at once. In this PR we take the appropriate cautious (potentially unnecessarily so) and incorporate gomodules as a major version increment in order to bypass strange problems with versioning.

To provide a complete proposal for merging this PR, here are some pros and cons.

Pros:
* This package is now in conformance to the official package management tool for the go ecosystem
* This resolves a perceived disincentive to adopt this library - the incompatibility with gomodules/
* This change is friendlier to users' development environments
* Gomod will help us enforce minimum supported versions of go

Cons:
* Users will need to include the major version in their import paths
* We need to be extra careful about unnecessary breaking changes
* We folded to 'the man' ;)

For the first con, consumers who are unhappy with this change should strongly consider wrapping their usage of `gofrs/uuid` in a UUID module so that they only have to update one import path when they upgrade this repository.

Additionally, for maintainers of this library, it is important for us to be aware of but not paralyzed by the impact of making breaking changes to the library. We have not made a breaking change in [2 years](https://github.com/gofrs/uuid/releases/tag/v4.0.0), and even though we're experimenting with v6 ad v7 UUIDs, there's nothing indicating that we are going to need to break the API anytime soon.

I'll leave this PR open for a week to collect feedback.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/gofrs/uuid/v5
+
+go 1.17


go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/gofrs/uuid/v5
+
+go 1.17
```
dylan-bourque, line 3: Not necessary, but we could go ahead and roll this forward to 1.19.

cameracker: Updated!

dylan-bourque (approved):
LGTM, but it would probably be a good idea to update the _Recommended Package Version_ section in `README.md` to point people at github.com/gofrs/uuid/v5 (with an explanation about **why** it's V5)

cameracker:
> 

Thanks Dylan, I'll look at doing that. I'll ask for a review again if I make that change

PR #117, opened 2023-02-15
Update to RFC 9562

### LINKED ISSUE #118: Link to new UUID draft
Hi! First of all, thanks for the library!

AFAIK the new UUID drafts didn't change the ~~uuidv6~~ / uuidv7 implementations, the README could be updated to say that this library is compatible with the most recent draft. WDYT?

Link to the new draft: https://datatracker.ietf.org/doc/html/draft-ietf-uuidrev-rfc4122bis-02

**Edit:** actually it looks like there are some changes to the uuidv6 implementation since the referenced draft in the readme.

kohenkatz:
~On 2022-10-17, the IETF uuidrev workgroup published a new draft that replaces peabody draft 04. On 2023-11-06, they released draft 14.~ On 2024-05-07, the IETF published RFC 9562, obsoleting RFC 4122.

This PR updates this project to implement the newest RFC. (fixes #118)

There are a few changes in documentation: mostly pointing to the new RFC URL, and adding comments as appropriate to explain the changes.

A new constant `VariantRFC9562` is added, with the old `VariantRFC4122` kept as an alias to it for backward compatibility.

The one major code change is in the implementation of UUID version 6. Since draft-01, this version now `SHOULD` use new pseudo-random data for the final 62 bytes, instead of using a `clock_seq` value for 14 of those bytes. (Although implementations `MAY` keep the old `clock_seq` behavior, this is discouraged, and is only allowed alongside keeping the MAC address as the `node` value. Since we are not keeping the MAC address, we also should not keep the `clock_seq`.


kohenkatz:
Sorry for the 13-month delay on this! I think it is ready for review now.

kohenkatz (commented):
I have updated this to match the published RFC, which is number 9562.

As far as I can tell, the only changes since the 14th draft have been grammatical, not technical.

cameracker (approved):

cameracker:
I just realized the coverage isnt running anymore...

cameracker:
Got codecov set back up but it probably wont run on this until we merge. I added @dylan-bourque  to the review to see if he has any feedback, but the changes seem correct to me and consistent with the verbiage in the latest RFC. Thank you for taking the time to update this.

dylan-bourque:
I gave this a quick scan and didn't see anything. I'm traveling today so won't have a chance to do a deeper review probably until Monday.

generator.go
```diff
@@ -272,92 +260,59 @@ func (g *Gen) NewV4() (UUID, error) {
...
 // pseudorandom data. The timestamp in a V6 UUID is the same as V1, with the bit
 // order being adjusted to allow the UUID to be k-sortable.
-//
-// This is implemented based on revision 03 of the Peabody UUID draft, and may
-// be subject to change pending further revisions. Until the final specification
-// revision is finished, changes required to implement updates to the spec will
-// not be considered a breaking change. They will happen as a minor version
-// releases until the spec is final.
 func (g *Gen) NewV6() (UUID, error) {
+	/* https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-6
```
dylan-bourque, line 272: Nit: maybe include the source link in the godoc comment so that someone doesn't have to dive into the code to find it.

generator.go
```diff
@@ -272,92 +260,59 @@ func (g *Gen) NewV4() (UUID, error) {
...
-	g.clockSequenceOnce.Do(func() {
-		buf := make([]byte, 2)
-		if _, err = io.ReadFull(g.rand, buf); err != nil {
-			return
-		}
-		g.clockSequence = binary.BigEndian.Uint16(buf)
-	})
-	if err != nil {
-		return 0, 0, err
+	//Based on the RFC 9562 recommendation that this data be fully random and not a monotonic counter,
```
dylan-bourque, line 295: Nit: comments that aren't directives should have a space after `//`

generator.go
```diff
@@ -272,92 +260,59 @@ func (g *Gen) NewV4() (UUID, error) {
...
-// This is implemented based on revision 04 of the Peabody UUID draft, and may
-// be subject to change pending further revisions. Until the final specification
-// revision is finished, changes required to implement updates to the spec will
-// not be considered a breaking change. They will happen as a minor version
-// releases until the spec is final.
 func (g *Gen) NewV7() (UUID, error) {
 	var u UUID
-	/* https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-04.html#name-uuid-version-7
-		0                   1                   2                   3
+	/* https://datatracker.ietf.org/doc/html/rfc9562#name-uuid-version-7
```
dylan-bourque, line 314: Same nit here

dylan-bourque (approved):
LGTM with a couple of nits

generator.go
cameracker:
```suggestion
	// Based on the RFC 9562 recommendation that this data be fully random and not a monotonic counter,
```

generator.go
kohenkatz: @dylan-bourque I thought about that, but I think it's too much information that most users won't care about. The link is only there to show the origin of the bit layout below it, which is not something that needs to be in the godoc. The README has links to the full RFC, and that shows up in the godoc already.

PR #129, opened 2023-11-13
add max uuid

### LINKED ISSUE #120: add max uuid with all bits set.
the opposite of the nil, its useful constant to have around.

i'll also note both should be better served by functions as global vars are mutable.

james-lawrence:
still think using function is better than var in this case.

fixes #120

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -161,8 +161,8 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 // zero.
 var Nil = UUID{}
 
-// Max is the maximum UUID, as specified in RFC-4122, that has all 128 bits
-// set to one.
+// Max is the maximum UUID, as specified in a draft of RFC-4122, that has all 128 bits
+// set to one. This feature is subject to removal if it is removed from the final draft of the RFC.
 var Max = UUID{
 	0xFF,
 	0xFF,


uuid.go
```diff
@@ -161,6 +161,27 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 // zero.
 var Nil = UUID{}
 
+// Max is the maximum UUID, as specified in RFC-4122, that has all 128 bits
+// set to one.
+var Max = UUID{
```
dylan-bourque, line 166: I think this should be called `Omni` to match the verbiage in the proposed update to the RFC.  This is not a "maximum" UUID value in the same way that `Nil` is not a "minimum" UUID.

The doc string should also include a callout that it's based on a proposed, but as yet unaccepted, revision and, as such, it is subject to change (or even be removed).

dylan-bourque: Just checked the latest draft update [here](https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/) and this actually is called "Max UUID" now.  I still personally disagree with the name because it's not accurate, we should be consistent with the RFC.

Still need the doc string callout, though.

uuid.go
```diff
@@ -161,6 +161,27 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 // zero.
 var Nil = UUID{}
 
+// Max is the maximum UUID, as specified in RFC-4122, that has all 128 bits
```
cameracker, line 164:
```suggestion
// Max is the maximum UUID, as specified in a draft of RFC-4122, that has all 128 bits
```

uuid.go
```diff
@@ -161,6 +161,27 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
...
+// set to one.
```
cameracker, line 165:
```suggestion
// set to one. This feature is subject to removal if it is removed from the final draft of the RFC.
```

uuid.go
cameracker: I went ahead and tweaked it. Hope it's not a faux pas :)

cameracker:
@james-lawrence , thank you very much for this contribution. I apologize it took me so long to merge it. I'll release it shortly :)

Issue #86, opened 2020-07-23
Support for Go 1.13 Error Wrapping


krak3n:
Hi

I was wondering if you had any interest in support for go1.13 error wrapping? For example:

``` go
const (
	ErrIncorrectLength Error = iota + 1
	// Other sentinel error types here
)

type Error int

func (e Error) Error() string {
	switch e {
	case ErrIncorrectLength:
		return "incorrect length"
	// Other error strings here
	default:
		return "unknown error"
	}
}

func (u *UUID) UnmarshalText(text []byte) error {
	switch len(text) {
	// Previous switch conditions
	default:
		return fmt.Errorf("uuid: %w %d in string %q", ErrIncorrectLength, len(text), text)
	}
}
```

This would allow for the usage of `errors.Is()` and `errors.As()`, useful for testing or performing other conditional logic based on the type of error, for example:

``` go
id, err := uuid.FromString("foo")
if err != nil {
	if errors.Is(err, uuid.ErrIncorrectLength) {
		// some bespoke logic
	}
}
```

If you approve I'd be more than happy to open a PR with the implementation detail.If backwards compatibility is a concern we can probably get around it with build tags.

Demo over on the playground: https://play.golang.org/p/H5teBSdvfct

jaddr2line:
We support go 1.13 error wrapping.

jaddr2line:
I copy-pasted your example into https://play.tinygo.org/ and it worked fine. Is there something wrong somewhere?

krak3n:
https://github.com/gofrs/uuid/blob/master/codec.go#L117 does not use error wrapping as my example above.

jaddr2line:
Wait sorry i mixed up what repo i was looking at oops.

krak3n:
So what do you think @jaddr2line is it worth me working in a PR for it?

krak3n:
@theckman indeed and ultimately if the end-user is just printing `err.Error()` then the underlying behaviour is no different, there error string presented to the user in my example above is the same as the one that exists currently. However, now that Go offers a way to wrap errors in error stings and return a wrapped sentinel error (as shown above) or a concrete struct type (like `os.PathError`) we can offer the end-user some more flexibility in how to handle specific types of errors.

[...]

PR #131, opened 2024-03-27
Implemented support for checkable errors

PatrLind:
This is a possible fix for issue https://github.com/gofrs/uuid/issues/86.

I have added two custom error types that can be used to check if an error is coming from the uuid package.
Not sure if the package maintainers agree with my implementation, but feel free to tell me if it is suitable or not.
Also not sure if there needs to be two errors types. Perhaps one is good enough?

No error messages should be different from before. `err.Error()` should return the same string.
It should now also be possible to check for uuid errors by using the `errors.Is()` and `errors.As()` functions.

I added a test function to verify that the implementation is OK.


PatrLind:
Well... I had a need for it and saw that no one had done anything to fix the bug that was reported 4 years ago.
My use case is that I have an error handler for gRPC that checks for some types of errors that can be returned. One of the errors I check is if the UUID could be parsed correctly. If it was not parsed correctly I send can send back the error to the user in clear text. Otherwise I just send a generic Internal Server Error. So this is a way to filter out internal errors that should not be sent back to the user while letting some specific errors thru. And since the UUID comes from the caller this is important information to know (that the parsing failed).
Right now I need to check if the error string contains "uuid: in" or "uuid: UUID must". This is a bit brittle, and it seems to me that looking for strings in an error message in order to determine if they are from the goofr/uuid package is not really a good practice.

In my experience most packages (at least the ones I use) export errors I can check against with errors.Is().

I'm not exactly sure what you mean by "any error returned by any function/method is, by definition, generated by this package since all of them are constructed directly"?

PatrLind:
Sure, declaring pre-allocated errors like in your example would have been great!
But since I'm pretty sure people have come to depend on the exact error strings in may ways I don't know about, so the current error messages cannot change (the string value returned from Error()). And since a lot of the error messages are including details on how the parsing failed, so those cannot be pre-allocated like that (right?).

I can refactor my code a bit and try to simplify it a bit.

dylan-bourque:
Thinking out loud ... we could probably use the "constant" sentinel errors *and* keep the current error strings with careful applications of `fmt.Errorf()`

Assuming `ErrInvalidFormat` is defined the way I mentioned before, the string returned by `.Error()` should be the same for the 2 expressions below
```
fmt.Errorf("uuid: incorrect UUID format in string %q", s)
fmt.Errorf("%w %q", ErrInvalidFormat, s)
```

error_test.go
```diff
@@ -45,10 +43,158 @@ func TestError(t *testing.T) {
...
+	if err == nil {
+		t.Error("expected an error")
+		return
+	}
+	if err.Error() != expectedErr {
+		t.Errorf("unexpected error '%s' != '%s'", err.Error(), expectedErr)
+	}
+
+	// no hw address error
+	netInterfaces = func() ([]net.Interface, error) {
```
dylan-bourque, line 137: This will potentially break parallel tests.  Would be better to move it to it's own test with a callout that it needs to modify the global state.

error.go
```diff
@@ -1,45 +1,22 @@
 package uuid
 
-import (
-	"fmt"
+type Error string
```
dylan-bourque, line 3: Need doc comments here and on the constants and methods below.

error.go
```diff
@@ -1,45 +1,22 @@
...
-
-func (e *ErrUUIDInvalidFormat) Unwrap() error {
-	return e.err
+func (e Error) Error() string {
+	return string(e)
 }
 
-func (e *ErrUUIDInvalidFormat) Is(target error) bool {
-	_, ok := target.(*ErrUUIDInvalidFormat)
+func (e Error) Is(target error) bool {
```
dylan-bourque, line 19: Is this explicit `Is()` method necessary?  I thought `errors.Is()` already includes this same logic and you only need the `Is()` method is you're doing something non-standard.

error_test.go
```diff
@@ -45,10 +43,158 @@ func TestError(t *testing.T) {
...
-			uuidErr := &ErrUUID{}
+			var uuidErr Error
 			if !errors.As(tc.err, &uuidErr) {
 				t.Error("expected errors.As() to work")
 			}
 		})
 	}
 }
+
+func TestAllErrorMessages(t *testing.T) {
```
dylan-bourque, line 54: This test is testing several different operations (`Parse()`, `Scan()`, `UnmarshalBinary()`, `FromStringOrNil()`, and several `TimestampFromV*` variants).  A given test should be testing one thing so this should be broken apart IMO.

error.go
PatrLind: I believe it is needed, the code doesn't work otherwise. But I can do some more research perhaps.

error_test.go
PatrLind: It's just testing the error messages, not the operations/functions themselves. So I think this is better.

error.go
dylan-bourque: You definitely don't need to provide `Is()` yourself.  [This playground](https://go.dev/play/p/jaCtGzLYIrS) shows that `errors.Is()` and `errors.As()` both work for plain and wrapped errors without needing explicit `Is()` or `As()` methods.

PatrLind: With the code you provided in the playground, no it's not needed. But with the code in my tests it's needed.
It is needed in order to check if the error returned is of type uuid.Error.
Like this: `if errors.Is(err, &fooError)`
Otherwise you can only user errors.As() and it is not always that you want to get the actual error object. Especially in this case since the error object doesn't contain any additional information. It is more interesting to know if the error is a uuid.Error type (doesn't matter if it is a ErrIncorrectFormatInString or ErrIncorrectLength or whatever...)
If you read the documentation of `errors.Is()` they give an example of this specific usage.

dylan-bourque: I see what's happening.  `errors.Is()` checks whether or not `err` is (or wraps) a *specific* sentinel error, not whether it is of a given type. Use `errors.As()` to check if an error is of a certain type.
```
// was the error specifically ErrInvalidFormat?
if errors.Is(err, uuid.ErrInvalidFormat) {
   ...
}
// is the concrete type of err uuid.Error?
// - note that this is uuid.Error, *not* *uuid.Error because the declared constants aren't pointers
var ue uuid.Error
if errors.As(err, &ue) {
   ...
}
```

PatrLind: Ok, so you don't think the `Is()` function should be provided, but the user should use `As()` instead?

PatrLind: Ok I see. I guess that's a valid point of view, but it is a bit different from how the standard lib does it. Here is an example:
```
func (e Errno) Is(target error) bool {
	switch target {
	case oserror.ErrPermission:
		return e == EACCES || e == EPERM
	case oserror.ErrExist:
		return e == EEXIST || e == ENOTEMPTY
	case oserror.ErrNotExist:
		return e == ENOENT
	case errorspkg.ErrUnsupported:
		return e == ENOSYS || e == ENOTSUP || e == EOPNOTSUPP
	}
	return false
}
```
Since I don't think I can convince you, and since it's easy to implement again in the future I guess I should remove it for now.

dylan-bourque: Correct. That `Is()` method is comparing `target` for equality against a specific, finite set of sentinel errors.  Nothing in that logic involves the concrete type of `target`.

error_test.go
```diff
@@ -4,42 +4,33 @@ import (
...
-			expectedTarget: ErrIncorrectFormatInString,
+			err:      fmt.Errorf("%w %q", ErrIncorrectFormatInString, "test"),
+			expected: "uuid: incorrect UUID format in string \"test\"",
 		},
 	}
 	for i, tc := range tcs {
 		t.Run(fmt.Sprintf("Test case %d", i), func(t *testing.T) {
 			var e2 Error
-			if !errors.Is(tc.err, &e2) {
+			if !errors.As(tc.err, &e2) {
```
dylan-bourque, line 31: This should still be testing that `errors.Is()` works correctly, it just needs to use the specific sentinel errors to do so.

I would add another `expectedErr Error` field to your test case struct, set to the wrapped sentinel error, and test it using something like this:
```
if !errors.Is(err, tc.expectedErr) {
    t.Errorf("expected error to be, or wrap, the %v sentinel error", tc.expectedErr)
}
```
As it's written right now we're testing `errors.As()` twice.

error_test.go
```diff
@@ -38,6 +42,9 @@ func TestIsAsError(t *testing.T) {
 			if !errors.As(tc.err, &uuidErr) {
```
dylan-bourque, line 42: This test is still duplicating the one just above.

dylan-bourque (approved):
LGTM.  Thanks for the submission and for working with me to resolve the feedback.

cameracker:
Hi @PatrLind , I'm finally getting some time to take a look at the open PRs. I decided to merge in the updated RFC one because it seemed like it'd have the least disruption if I merged it in first. Apologies if this generates conflicts. Also, codecov hasnt been running for a bit so the next time a change is made to this review it may come up with some missing coverage. I'll only ask for test coverage on areas you introduced.

cameracker (approved):

Issue #132, opened 2024-04-08
Wrong error message for TimestampFromV7


PatrLind:
Seems like the UUID library returns the wrong error message for function `TimestampFromV7`.
The returned error message is (with input 'e86160d3-beff-443c-b9b5-1f8197ccb12e':
`uuid: e86160d3-beff-443c-b9b5-1f8197ccb12e is version 4, not version 6`
The expected error message should be:
`uuid: e86160d3-beff-443c-b9b5-1f8197ccb12e is version 4, not version 7`

cameracker:
Hi, this has been closed with #127

PR #127, opened 2023-10-30
Fix TimestampFromV7 error message

jaredLunde:
6 -> 7

PR #133, opened 2024-04-08
Fix for wrong version in error message

PatrLind:
This fixes that the error message returns the wrong version nr. This is a possible solution to my issue https://github.com/gofrs/uuid/issues/132


PatrLind:
Oh, I just saw that there is another PR for the exact same thing!
https://github.com/gofrs/uuid/pull/127

I guess this one can be removed? Or why isn't the other one merged yet?

cameracker:
Thanks, I'm going to close this in favor of the other. Really appreciate the catch :)

Issue #147, opened 2024-05-13
Tune up Fuzz test


cameracker:
Several years ago, @acln0  contributed a fuzz test to the package It was incorporated prior to github actions, and was not integrated into travisCI (what we were using at the time), so it hasn't made it into our CI. Additionally, the way it's invoked involves installing `dvyukov/go-fuzz` using `go get` - it's not included in `go.mod`. And, go provides fuzzing functionality directly as of Go 1.18.

We should touch this up. Fuzzing is a valuable testing practice, and even if we don't find much because our parameterized functions are so minimal, it'd still be useful to cover our bases.

I suggest fuzz testing should look like
* We have fuzz testing for our string parsing functions (implemented today)
* We have fuzz testing for any functions that take user input, unless its a complicated input (like passing a custom generator)
~* We run this fuzz test for 30 seconds during CI~
* We use the go fuzzing functionality, move off of `go-fuzz` 

Interested in with others think!

dylan-bourque:
The only thing I would push back on is running active fuzzing (generating new input values) during CI.  We should definitely run active fuzzing and commit any edge cases it finds alongside the fixes, but no one likes it when their CI build fails for something unrelated to their change.

cameracker:
Fixed in #148

PR #148, opened 2024-05-14
Update fuzz tests to use go fuzz features

cameracker:
This PR implements #147. Now, we use the fuzz test features built into `go test`.

This PR also replaces the `go-fuzz` tests and their accompanying corpus.


codec_test.go
```diff
@@ -403,28 +400,104 @@ func BenchmarkParseV4(b *testing.B) {
...
+	for _, seed := range fromStringCorpus {
+		f.Add(seed)
+	}
+	uuidRegexp, err := regexp.Compile(uuidPattern)
+	if err != nil {
+		f.Error("uuid regexp failed to compile")
 	}
+	f.Fuzz(func(t *testing.T, payload string) {
+		u, err := FromString(payload)
+		if err != nil {
```
cameracker, line 476: the general logic for this pair is that if the function didn't return an error, we make sure the uuid matches a regex for uuids. Otherwise, we assume the error was correct for the input. This implicitly fails the fuzz test on a panic.

codec_test.go
```diff
@@ -403,28 +400,104 @@ func BenchmarkParseV4(b *testing.B) {
...
-		return os.WriteFile(path, []byte(data), os.ModePerm)
+	uuidRegexp, err := regexp.Compile(uuidPattern)
+	if err != nil {
+		f.Error("uuid regexp failed to compile")
 	}
-	for _, fst := range fromStringTests {
-		name := "seed_valid_" + fst.variant
-		if err := writeSeedFile(name, fst.input); err != nil {
-			t.Fatal(err)
+	f.Fuzz(func(t *testing.T, payload []byte) {
```
cameracker, line 419: the general logic for this pair is that if the payload was **not** 16 bytes, we fail the fuzz if fan error was **not** returned (or put another way, we verify that the function returns an error when the payload is 16 bytes). Otherwise, if an error was not returned, we make sure it's not a Nil uuid, and we make sure it matched the uuid pattern

.github/workflows/codeql.yml
```diff
@@ -40,11 +40,6 @@ jobs:
         # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
 
     steps:
-      - name: Harden Runner
```
cameracker, line 43: Removed harden in places that I missed it.

codec_test.go
```diff
@@ -403,28 +400,109 @@ func BenchmarkParseV4(b *testing.B) {
 	}
 }
 
-var seedFuzzCorpus = flag.Bool("seed_fuzz_corpus", false, "seed fuzz test corpus")
+const uuidPattern = "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
```
dylan-bourque, line 403: Should this also include capital `A-F` in the match patterns since those would still be valid UUID strings?

cameracker, line 403:
```suggestion
const uuidPattern = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
```

codec_test.go
```diff
@@ -403,28 +400,109 @@ func BenchmarkParseV4(b *testing.B) {
...
+}
+
+var fromStringCorpus = []string{
+	"6ba7b810-9dad-11d1-80b4-00c04fd430c8",
+	"{6ba7b810-9dad-11d1-80b4-00c04fd430c8}",
+	"urn:uuid:6ba7b810-9dad-11d1-80b4-00c04fd430c8",
+	"6ba7b8109dad11d180b400c04fd430c8",
+	"{6ba7b8109dad11d180b400c04fd430c8}",
+	"urn:uuid:6ba7b8109dad11d180b400c04fd430c8",
+}
```
cameracker, line 465:
```suggestion
var fromStringCorpus = []string{
	"6ba7b810-9dad-11d1-80b4-00c04fd430c8",
	"6BA7B810-9DAD-11D1-80B4-00C04FD430C8",
	"{6BA7B810-9DAD-11D1-80B4-00C04FD430C8}",
	"urn:uuid:6BA7B810-9DAD-11D1-80B4-00C04FD430C8",
	"6BA7B8109DAD11D180B400C04FD430C8",
	"{6BA7B8109DAD11D180B400C04FD430C8}",
	"urn:uuid:6BA7B8109DAD11D180B400C04FD430C8",
}
```

codec_test.go
cameracker: Thanks, nice catch

.github/workflows/go.yml
```diff
@@ -38,20 +32,16 @@ jobs:
       run: go test ./... -race -coverprofile=coverage.txt -covermode=atomic
 
     - name: Coverage
-      uses: codecov/codecov-action@125fc84a9a348dbcf27191600683ec096ec9021c # v4.4.1
+      uses: codecov/codecov-action@125fc84a9a348dbcf27191600683ec096ec9021c # v4.3.1
```
dylan-bourque, line 35: Was this downgrade intentional?  Looks like the new hash is mapping to an older release.

codec_test.go
```diff
@@ -403,28 +400,110 @@ func BenchmarkParseV4(b *testing.B) {
...
+	for _, seed := range fromBytesCorpus {
+		f.Add(seed)
 	}
-	corpusDir := filepath.Join(".", "testdata", "corpus")
-	writeSeedFile := func(name, data string) error {
-		path := filepath.Join(corpusDir, name)
-		return os.WriteFile(path, []byte(data), os.ModePerm)
+	uuidRegexp, err := regexp.Compile(uuidPattern)
+	if err != nil {
+		f.Error("uuid regexp failed to compile")
```
dylan-bourque, line 417: I think this should be a fatal error since test test can't function if `uuidPattern` isn't valid.  I'd even consider using `regexp.MustCompile()` and let it panic.

codec_test.go
```diff
@@ -403,28 +400,110 @@ func BenchmarkParseV4(b *testing.B) {
...
-			t.Fatal(err)
+	f.Fuzz(func(t *testing.T, payload []byte) {
+		u, err := FromBytes(payload)
+		if len(payload) != Size && err == nil {
+			t.Errorf("%v did not result in an error", payload)
+		}
+		if len(payload) == Size && u == Nil {
+			t.Errorf("%v resulted in Nil uuid", payload)
 		}
+		if len(payload) == Size && !uuidRegexp.MatchString(u.String()) {
```
dylan-bourque, line 427: Can this `MatchString()` call actually ever fail, i.e. will `u.String()` ever return a string that won't match?

dylan-bourque (commented):
Had a few questions/comments but I like the change. 👍

.github/workflows/go.yml
cameracker: Ah, no, I messed up the rebase on accident. Thanks for the catch.

.github/workflows/go.yml
```diff
@@ -39,19 +33,15 @@ jobs:
 
     - name: Coverage
       uses: codecov/codecov-action@125fc84a9a348dbcf27191600683ec096ec9021c # v4.4.1
```
cameracker, line 35:
```suggestion
      uses: codecov/codecov-action@e28ff129e5465c2c0dcc6f003fc735cb6ae0c673 # v4.5.0
```

codec_test.go
dylan-bourque: 👍

dylan-bourque (approved):
LGTM

Issue #195, opened 2025-02-08
TimestampFromUUIDv7 Artificially Limited


nathanmcgarvey-modopayments:
### Welcome

- [x] Yes, I'm using the latest release.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.

### What did you expect to see?

The maximum UUIDv7 should have a timestamp that evaluates to the year 10889 per https://www.rfc-editor.org/rfc/rfc9562.html#section-6.1-2.8

> Length:
>     The length of a given timestamp directly impacts how many timestamp ticks can be contained in a UUID before the maximum value for the timestamp field is reached. Take care to ensure that the proper length is selected for a given timestamp. UUIDv1 and UUIDv6 utilize a 60-bit timestamp valid until 5623 AD; UUIDv7 features a 48-bit timestamp valid until the year 10889 AD.

### What did you see instead?

The maximum UUIDv7 had a timestamp that evaluated to the year 2121. 

```
go test -v -timeout 30s -run ^TestTimestampFromV7$ github.com/gofrs/uuid/v5 -race -count=1 -v

=== RUN   TestTimestampFromV7
    uuid_test.go:275: 1969-12-31 18:00:00 -0600 CST <nil>
    uuid_test.go:275: 2023-09-13 14:02:18.605 -0500 CDT <nil>
    uuid_test.go:275: 2121-04-11 06:53:25.0117257 -0500 CDT <nil>
--- PASS: TestTimestampFromV7 (0.00s)
PASS
ok  	github.com/gofrs/uuid/v5	1.194s
```

This is likely because of inappropriate function re-use between v1 and v7 and an overflow somewhere (though I haven't fully debugged to prove that):

uuid.go:84-67
```
// Timestamp is the count of 100-nanosecond intervals since 00:00:00.00,
// 15 October 1582 within a V1 UUID. This type has no meaning for other
// UUID versions since they don't have an embedded timestamp.
type Timestamp uint64

const _100nsPerSecond = 10000000

// Time returns the UTC time.Time representation of a Timestamp
func (t Timestamp) Time() (time.Time, error) {
	secs := uint64(t) / _100nsPerSecond
	nsecs := 100 * (uint64(t) % _100nsPerSecond)

	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)), nil
}
```

### Reproduction steps

```console
Simple patch to the existing test file to see the values embedded in already-present tests:


[...]

Issue #128, opened 2023-10-30
TimestampFromV7 should return the unix timestamp in milliseconds


jaredLunde:
Apologies if I am missing something, but it appears the current implementation of `TimestampFromV7()` uses the Gregorian epoch timestamp as opposed to unix: 

https://github.com/gofrs/uuid/blob/22c52c268bc0dcc0569793f5b1433db423f5a9c6/uuid.go#L156
https://github.com/gofrs/uuid/blob/22c52c268bc0dcc0569793f5b1433db423f5a9c6/generator.go#L37-L39

The draft states:
> unix_ts_ms:
> 48 bit big-endian unsigned number of Unix epoch timestamp as per Section 6.1.

So shouldn't this be:
```go
func TimestampFromV7(u UUID) (Timestamp, error) {
	if u.Version() != 7 {
		return 0, fmt.Errorf("uuid: %s is version %d, not version 6", u, u.Version())
	}

	t := 0 |
		(int64(u[0]) << 40) |
		(int64(u[1]) << 32) |
		(int64(u[2]) << 24) |
		(int64(u[3]) << 16) |
		(int64(u[4]) << 8) |
		int64(u[5])
        
	return Timestamp(t), nil
}
```

Where tests would be as straightforward as:
```go
	tests := []struct {
		u       UUID
		want    Timestamp
		wanterr bool
	}{
		{u: Must(FromString("00000000-0000-7000-0000-000000000000")), want: Timestamp(0x000000000000)},
		{u: Must(FromString("018a8fec-3ced-7164-995f-93c80cbdc575")), want: Timestamp(0x018a8fec3ced)},
		{u: Must(FromString("ffffffff-ffff-7fff-ffff-ffffffffffff")), want: Timestamp(0xffffffffffff)},
       }
```

dylan-bourque:
I didn't write the existing implementation, but it looks like it's reusing the `Timestamp` type and adjusting the value to match that type's documented constraints [here](https://github.com/gofrs/uuid/blob/master/uuid.go#L84-L87).

jaredLunde:
Ahh, should we expect each parser to return its draft-compliant timestamp though? Happy to draft that PR if you think so!

dylan-bourque:
That would imply creating a separate `TimestampVN` type for each version, which doesn't feel like a good path. 🤔

jaredLunde:
Or separate by epoch
- `UnixTime`, `GregTime`
[...]

Issue #81, opened 2019-08-07
Adding the ability to encode/decode to/from Base64 and Base32


asahaf:
I've found myself encoding the UUID to Base64 and Base32 string in multiple projects I worked on. 
Do you think it's a good idea if we add this encoding and decoding to this library or that's out of its scope?
Thanks

zerkms:
What's the use case? I personally never seen it encoded anything but binary or hex-string.

asahaf:
One use case is when I use UUIDs in URLs. Base64 produces shorter string representation. in case I need shorter case-insensitive UUID string I use Base32.

cameracker:
Equivalent functionality is https://www.npmjs.com/package/short-uuid

dylan-bourque:
Personally, I'd argue that this would be an unnecessary extension, since it would essentially be 1-line pass-through calls to `base64.UrlEncoding.EncodeToString(u.Bytes())` and `base32.StdEncoding.EncodeToString(u.Bytes())`.

wolfeidau:
I have a wrapper which uses base58, a handy middle ground which is used widely https://github.com/wolfeidau/shortuuid, as you say it is a pretty simple wrapper around existing UUID libraries. 

Keen to migrate this to this library so i can try out v6 UUIDs.

cameracker:
Doing a little research on the topic, I think we should use base58 as is done with @wolfeidau's package. 

> Similar to Base64, but modified to avoid both non-alphanumeric characters (+ and /) and letters that might look ambiguous when printed (0 – zero, I – capital i, O – capital o and l – lower-case L). Base58 is used to represent [bitcoin](https://en.wikipedia.org/wiki/Bitcoin) addresses.[[citation needed](https://en.wikipedia.org/wiki/Wikipedia:Citation_needed)] Some messaging and social media systems [break lines](https://en.wikipedia.org/wiki/Line_wrap_and_word_wrap) on non-alphanumeric strings. This is avoided by not using [URI reserved characters](https://en.wikipedia.org/wiki/Percent-encoding#Reserved_characters) such as +. For [SegWit](https://en.wikipedia.org/wiki/SegWit), it was replaced by Bech32, see below.

Base58 in the original bitcoin source code

https://en.wikipedia.org/wiki/Binary-to-text_encoding#Base58

dylan-bourque:
[...]

PR #197, opened 2025-02-08
Fix Comment on the the Timestamp.Time() method

### LINKED ISSUE #196: Timestamp.Time() not returning UTC
### Welcome

- [x] Yes, I'm using the latest release.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.

### What did you expect to see?

The return from the Timestamp.Time() function should be UTC per it's own function comment:

> // Time returns the UTC time.Time representation of a Timestamp

Likely fix: uuid.go line 96 is missing a .UTC() on the end

### What did you see instead?

Timezones returned from the Timestamp.Time() function are undefined or local system time or something else entirely (not sure why it varies between executions). 

### Reproduction steps

```console
Simple patch to the existing test file to see the values embedded in already-present tests:


diff --git a/uuid_test.go b/uuid_test.go
index 6dc97f9..7837c45 100644
--- a/uuid_test.go
+++ b/uuid_test.go
@@ -271,6 +271,10 @@ func TestTimestampFromV7(t *testing.T) {
        for _, tt := range tests {
                got, err := TimestampFromV7(tt.u)
 
+               if err == nil {
+                       t.Log(got.Time())
+               }
+
                switch {
                case tt.wanterr && err == nil:
                        t.Errorf("TimestampFromV7(%v) want error, got %v", tt.u, got)




go test -v -timeout 30s -run ^TestTimestampFromV7$ github.com/gofrs/uuid/v5 -race -count=1 -v

=== RUN   TestTimestampFromV7
    uuid_test.go:275: 1969-12-31 18:00:00 -0600 CST <nil>
    uuid_test.go:275: 2023-09-13 14:02:18.605 -0500 CDT <nil>
    uuid_test.go:275: 2121-04-11 06:53:25.0117257 -0500 CDT <nil>
--- PASS: TestTimestampFromV7 (0.00s)
PASS
ok  	github.com/gofrs/uuid/v5	1.194s
```

### Version of flock

N/A

### Logs

```console
N/A
```

### Go environment

go version go1.23.3 darwin/arm64

### Validation

- [x] Yes, I've included all information above (version, config, etc.).

nathanmcgarvey-modopayments:
Fixes #196

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -93,7 +93,7 @@ func (t Timestamp) Time() (time.Time, error) {
 	secs := uint64(t) / _100nsPerSecond
 	nsecs := 100 * (uint64(t) % _100nsPerSecond)
 
-	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)), nil
+	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)).UTC(), nil
 }
 
 // TimestampFromV1 returns the Timestamp embedded within a V1 UUID.
--- a/uuid.go
+++ b/uuid.go
@@ -88,12 +88,16 @@ type Timestamp uint64
 
 const _100nsPerSecond = 10000000
 
-// Time returns the UTC time.Time representation of a Timestamp
+// Time returns the time.Time representation of a Timestamp.
+//
+// The resulting time.Time:
+// - Contains only wall clock time (no monotonic clock component)
+// - Uses the local system timezone for the location
 func (t Timestamp) Time() (time.Time, error) {
 	secs := uint64(t) / _100nsPerSecond
 	nsecs := 100 * (uint64(t) % _100nsPerSecond)
 
-	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)).UTC(), nil
+	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)), nil
 }
 
 // TimestampFromV1 returns the Timestamp embedded within a V1 UUID.


nathanmcgarvey-modopayments:
Fixes: #196

kohenkatz (approved):
I think this makes sense, but I don't want to merge it without at least one other approval.

kohenkatz:
I thought of one situation in which this could be a breaking change. From the documentation for time.Time:

>Note that the Go == operator compares not just the time instant but also the Location and the monotonic clock reading. Therefore, Time values should not be used as map or database keys without first guaranteeing that the identical Location has been set for all values, which can be achieved through use of the UTC or Local method, and that the monotonic clock reading has been stripped by setting t = t.Round(0).

So if anyone was relying on the previous behavior of being Local, this will be a problem for them.

Is this something we need to worry about?

ldez:
Maybe you can do something like that:

```go

// Time returns the time.Time representation of a Timestamp.
// Deprecated: use TimeUTC instead.
func (t Timestamp) Time() (time.Time, error) {
	secs := uint64(t) / _100nsPerSecond
	nsecs := 100 * (uint64(t) % _100nsPerSecond)

	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)), nil
}

// TimeUTC returns the UTC time.Time representation of a Timestamp.
func (t Timestamp) TimeUTC() (time.Time, error) {
	base, err := t.Time()
	if err != nil {
		return time.Time{}, err
	}

	return base.UTC(), nil
}

```

nathanmcgarvey-modopayments:
From my very small perspective in my very small corner of the world:

The current behavior *is* broken, both in theory and practice.

In practice:

```golang
func TestTimeZone(t *testing.T) {
	uuids := []UUID{
		Must(FromString("00000000-0000-7000-8000-000000000000")),
		Must(FromString("10000000-0000-7000-8000-000000000000")),
		Must(FromString("20000000-0000-7000-8000-000000000000")),
		Must(FromString("30000000-0000-7000-8000-000000000000")),
		Must(FromString("30000000-0000-7000-8000-000000000000")),
		Must(FromString("40000000-0000-7000-8000-000000000000")),
		Must(FromString("50000000-0000-7000-8000-000000000000")),
		Must(FromString("60000000-0000-7000-8000-000000000000")),
		Must(FromString("70000000-0000-7000-8000-000000000000")),
		Must(FromString("80000000-0000-7000-8000-000000000000")),
		Must(FromString("90000000-0000-7000-8000-000000000000")),
		Must(FromString("a0000000-0000-7000-8000-000000000000")),
		Must(FromString("b0000000-0000-7000-8000-000000000000")),
		Must(FromString("c0000000-0000-7000-8000-000000000000")),
		Must(FromString("d0000000-0000-7000-8000-000000000000")),
		Must(FromString("e0000000-0000-7000-8000-000000000000")),
		Must(FromString("f0000000-0000-7000-8000-000000000000")),
	}

	for _, uid := range uuids {
		t.Run(uid.String(), func(t *testing.T) {
			got, _ := TimestampFromV7(uid)
			tm, err := got.Time()
			t.Log(tm.Zone())
			if err != nil {
				t.Fatalf("uuid.Time(%q) failed: %v", uid, err)
			}
		})
	}
}
```

Output (at least for me) gives four time zones (CST,CWT,LMT,CDT):
```
go test -v -timeout 30s -run ^TestTimeZone$ github.com/gofrs/uuid/v5 -count=1 
=== RUN   TestTimeZone
=== RUN   TestTimeZone/00000000-0000-7000-8000-000000000000
    uuid_test.go:312: CST -21600
=== RUN   TestTimeZone/10000000-0000-7000-8000-000000000000
    uuid_test.go:312: CWT -18000
=== RUN   TestTimeZone/20000000-0000-7000-8000-000000000000
    uuid_test.go:312: CST -21600
=== RUN   TestTimeZone/30000000-0000-7000-8000-000000000000
    uuid_test.go:312: CST -21600
=== RUN   TestTimeZone/30000000-0000-7000-8000-000000000000#01
    uuid_test.go:312: CST -21600
=== RUN   TestTimeZone/40000000-0000-7000-8000-000000000000
    uuid_test.go:312: LMT -21036
=== RUN   TestTimeZone/50000000-0000-7000-8000-000000000000
    uuid_test.go:312: LMT -21036
=== RUN   TestTimeZone/60000000-0000-7000-8000-000000000000
    uuid_test.go:312: LMT -21036
=== RUN   TestTimeZone/70000000-0000-7000-8000-000000000000
    uuid_test.go:312: LMT -21036
=== RUN   TestTimeZone/80000000-0000-7000-8000-000000000000
    uuid_test.go:312: LMT -21036
=== RUN   TestTimeZone/90000000-0000-7000-8000-000000000000
    uuid_test.go:312: LMT -21036
=== RUN   TestTimeZone/a0000000-0000-7000-8000-000000000000
    uuid_test.go:312: LMT -21036
=== RUN   TestTimeZone/b0000000-0000-7000-8000-000000000000
    uuid_test.go:312: CDT -18000
=== RUN   TestTimeZone/c0000000-0000-7000-8000-000000000000
    uuid_test.go:312: CDT -18000
=== RUN   TestTimeZone/d0000000-0000-7000-8000-000000000000
    uuid_test.go:312: CDT -18000
=== RUN   TestTimeZone/e0000000-0000-7000-8000-000000000000
    uuid_test.go:312: CDT -18000
=== RUN   TestTimeZone/f0000000-0000-7000-8000-000000000000
    uuid_test.go:312: CDT -18000
--- PASS: TestTimeZone (0.00s)
    --- PASS: TestTimeZone/00000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/10000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/20000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/30000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/30000000-0000-7000-8000-000000000000#01 (0.00s)
    --- PASS: TestTimeZone/40000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/50000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/60000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/70000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/80000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/90000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/a0000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/b0000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/c0000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/d0000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/e0000000-0000-7000-8000-000000000000 (0.00s)
    --- PASS: TestTimeZone/f0000000-0000-7000-8000-000000000000 (0.00s)
PASS
ok  	github.com/gofrs/uuid/v5	0.170s
```

In theory:

As your quote from the time package points out:

> Time values should not be used as map or database keys without first guaranteeing that the identical Location has been set for all values, which can be achieved through use of the UTC or Local method, and that the monotonic clock reading has been stripped by setting t = t.Round(0)

The code currently does none of those (.UTC() or.Local(), and .Round(0)). So we're sure that the monotonic clock is changing regarding map keys. (A caller would have to be doing that themself, already.) And regarding equals and a timezone the time package provides a .Equal() method intentionally specifying the following:

> Equal reports whether t and u represent the same time instant. Two times can be equal even if they are in different locations. For example, 6:00 +0200 and 4:00 UTC are Equal. See the documentation on the Time type for the pitfalls of using == with Time values; most code should use Equal instead.



Additionally, this is memory-internal. Unless the caller is doing string or binary-equivalent comparison on a previously saved time.Time object E.g. intentionally saving the timezone off somewhere, then bringing it back and comparing, this is a non-issue. If only in memory, then when the process is restarted after updating this package, all the comparisons should work as usual. If they saved it off as a string or binary representation, that has no time zone data, either, so marshaling it back to a UUID should be fine. Only if someone is calling the .Time() method and then saving off some representation of it with timezone that persists process restarts or communicates with a system that includes the timezone, but is not actually parsing the timezone would this be an issue. (I think.... maybe I'm missing a corner case somewhere.)

kohenkatz:
> CST,CWT,LMT,CDT

This is actually working as expected. The four time zone abbreviations you are seeing are an artifact of the `tzdata` time zone database when dealing with past timestamps.

- CST - "Central Standard Time"
- CWT - "Central War Time" - used for WWII
- LMT - "Local Mean Time" - used pre-Standardized-Timezones
- CDT - "Central Daylight Time"

Your sample times are missing one other option in your area: "CPT" = "Central Peace Time" (after the conclusion of WWII).

The reason you are seeing this in an unexpected order ("LMT" seen *after* Standard time) is because of your other issue, #196, which is causing the dates to be calculated incorrectly.

Here is what I get (US Eastern time) if I add `t.Log(tm)` above `t.Log(tm.Zone())` in your example:

Note the use of "EWT" for 1942, "EST/EDT" for everything else after 1843 (the creation of "Railway Time" which became "Standard Time", and "LMT" for everything before that.

```
go.exe test -timeout 30s -run ^TestTimeZone$ github.com/gofrs/uuid/v5

=== RUN   TestTimeZone
=== RUN   TestTimeZone/00000000-0000-7000-8000-000000000000
    codec_test.go:545: 1969-12-31 19:00:00 -0500 EST
    codec_test.go:546: EST -18000
--- PASS: TestTimeZone/00000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/10000000-0000-7000-8000-000000000000
    codec_test.go:545: 1942-12-03 02:46:10.7064484 -0400 EWT
    codec_test.go:546: EWT -14400
--- PASS: TestTimeZone/10000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/20000000-0000-7000-8000-000000000000
    codec_test.go:545: 1915-11-04 08:32:21.4128968 -0500 EST
    codec_test.go:546: EST -18000
--- PASS: TestTimeZone/20000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/30000000-0000-7000-8000-000000000000
    codec_test.go:545: 1888-10-04 15:18:32.1193452 -0500 EST
    codec_test.go:546: EST -18000
--- PASS: TestTimeZone/30000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/30000000-0000-7000-8000-000000000000#01
    codec_test.go:545: 1888-10-04 15:18:32.1193452 -0500 EST
    codec_test.go:546: EST -18000
--- PASS: TestTimeZone/30000000-0000-7000-8000-000000000000#01 (0.00s)
=== RUN   TestTimeZone/40000000-0000-7000-8000-000000000000
    codec_test.go:545: 1861-09-05 22:08:40.8257936 -0456 LMT
    codec_test.go:546: LMT -17762
--- PASS: TestTimeZone/40000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/50000000-0000-7000-8000-000000000000
    codec_test.go:545: 1834-08-08 04:54:51.532242 -0456 LMT
    codec_test.go:546: LMT -17762
--- PASS: TestTimeZone/50000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/60000000-0000-7000-8000-000000000000
    codec_test.go:545: 1807-07-10 11:41:02.2386904 -0456 LMT
    codec_test.go:546: LMT -17762
--- PASS: TestTimeZone/60000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/70000000-0000-7000-8000-000000000000
    codec_test.go:545: 1780-06-09 18:27:12.9451387 -0456 LMT
    codec_test.go:546: LMT -17762
--- PASS: TestTimeZone/70000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/80000000-0000-7000-8000-000000000000
    codec_test.go:545: 1753-05-12 01:13:23.6515871 -0456 LMT
    codec_test.go:546: LMT -17762
--- PASS: TestTimeZone/80000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/90000000-0000-7000-8000-000000000000
    codec_test.go:545: 1726-04-13 07:59:34.3580355 -0456 LMT
    codec_test.go:546: LMT -17762
--- PASS: TestTimeZone/90000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/a0000000-0000-7000-8000-000000000000
    codec_test.go:545: 1699-03-14 14:45:45.0644839 -0456 LMT
    codec_test.go:546: LMT -17762
--- PASS: TestTimeZone/a0000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/b0000000-0000-7000-8000-000000000000
    codec_test.go:545: 2256-09-03 22:02:31.4804838 -0400 EDT
    codec_test.go:546: EDT -14400
--- PASS: TestTimeZone/b0000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/c0000000-0000-7000-8000-000000000000
    codec_test.go:545: 2229-08-06 04:48:42.1869322 -0400 EDT
    codec_test.go:546: EDT -14400
--- PASS: TestTimeZone/c0000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/d0000000-0000-7000-8000-000000000000
    codec_test.go:545: 2202-07-08 11:34:52.8933806 -0400 EDT
    codec_test.go:546: EDT -14400
--- PASS: TestTimeZone/d0000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/e0000000-0000-7000-8000-000000000000
    codec_test.go:545: 2175-06-08 18:21:03.5998289 -0400 EDT
    codec_test.go:546: EDT -14400
--- PASS: TestTimeZone/e0000000-0000-7000-8000-000000000000 (0.00s)
=== RUN   TestTimeZone/f0000000-0000-7000-8000-000000000000
    codec_test.go:545: 2148-05-10 01:07:14.3062773 -0400 EDT
    codec_test.go:546: EDT -14400
--- PASS: TestTimeZone/f0000000-0000-7000-8000-000000000000 (0.00s)
--- PASS: TestTimeZone (0.00s)
PASS
ok      github.com/gofrs/uuid/v5        0.502s
```

---

> The code currently does none of those (.UTC() or.Local(), and .Round(0))

Not explicitly, but it does the same thing in a not-well-documented way. Calls to `time.Unix` explicitly use the `Local` timezone, and already don't have any monotonic clock reading that would need to be stripped (by calling `Round(0)`).

> Additionally, this is memory-internal. Unless the caller is doing string or binary-equivalent comparison on a previously saved time.Time object ... (I think.... maybe I'm missing a corner case somewhere.)

The case you are missing here is exactly what is mentioned in the documentation: `map[time.Time]any` - the runtime will compare the time keys using `==`, not `.Equal()`. With the current implementation's implicit use of `Local`, a `time.Time` constructed in the local timezone pointing to the same timestamp would match. Changing this implementation to `UTC` would definitely cause that to break.

nathanmcgarvey-modopayments:
> > CST,CWT,LMT,CDT
> 
> This is actually working as expected. The four time zone abbreviations you are seeing are an artifact of dealing with past timestamps.
> 
>     * CST - "Central Standard Time"
> 
>     * CWT - "Central War Time" - used for WWII
> 
>     * LMT - "Local Mean Time" - used pre-Standardized-Timezones
> 
>     * CDT - "Central Daylight Time"
> 
> 
> Your sample times are missing one other option in your area: "CPT" = "Central Peace Time" (after the conclusion of WWII).
> 
> The reason you are seeing this in an unexpected order ("LMT" seen _after_ Standard time) is because of your other issue, #195, which is causing the dates to be calculated incorrectly.
> 
> Here is what I get (US Eastern time) if I add `t.Log(tm)` above `t.Log(tm.Zone())` in your example:
> 
> Note the use of "EWT" for 1942, "EST/EDT" for everything else after 1883 (the creation of "Railway Time" which became "Standard Time", and "LMT" for everything before that.


Well, now I learned something today! I'm calling that a good day.


> > The code currently does none of those (.UTC() or.Local(), and .Round(0))
> 
> Not explicitly, but it does the same thing in a not-well-documented way. Calls to `time.Unix` explicitly use the `Local` timezone, and already don't have any monotonic clock reading that would need to be stripped (by calling `Round(0)`).


Thanks for pointing that out. My quick read didn't catch that.


> The case you are missing here is exactly what is mentioned in the documentation: `map[time.Time]any` - the runtime will compare the time keys using `==`, not `.Equal()`. With the current implementation's implicit use of `Local`, a `time.Time` constructed in the local timezone pointing to the same timestamp would match. Changing this implementation to `UTC` would definitely cause that to break.

Yes, that makes sense. I pity the person attempting to wrangle a Local timezone as a map key. That sounds unpleasant.

nathanmcgarvey-modopayments:
> Maybe you can do something like that:
> 
> ```go
> // Time returns the time.Time representation of a Timestamp.
> // Deprecated: use TimeUTC instead.
> func (t Timestamp) Time() (time.Time, error) {
> 	secs := uint64(t) / _100nsPerSecond
> 	nsecs := 100 * (uint64(t) % _100nsPerSecond)
> 
> 	return time.Unix(int64(secs)-(epochStart/_100nsPerSecond), int64(nsecs)), nil
> }
> 
> // TimeUTC returns the UTC time.Time representation of a Timestamp.
> func (t Timestamp) TimeUTC() (time.Time, error) {
> 	base, err := t.Time()
> 	if err != nil {
> 		return time.Time{}, err
> 	}
> 
> 	return base.UTC(), nil
> }
> ```
> 
> This avoids any side effects or breaking changes.
> 
> The other safest approach is to only update the documentation:
> 
> ```go
> // Time returns the local time.Time representation of a Timestamp.
> ```
> 
> No possible regression but the existing "bug" will still be here if someone created its code based on the comment.

I like either of these.... why does Time() return an error? (I'll assume legacy reasons.)

kohenkatz:
> why does Time() return an error? (I'll assume legacy reasons.)

Not for legacy reasons. It is "for potential unknown cases in future to avoid [breaking changes]". 😐

See https://github.com/gofrs/uuid/pull/31#discussion_r211767224

nathanmcgarvey-modopayments:
Happy to update this PR. Is there a preference from the new function and deprecated function approach vs just updating the comment?

kohenkatz:
Personally, I would go with just updating the comment. Anyone who wants to call `.UTC()` on the returned `time.Time` instance can do so easily. As we've previously discussed in #81, #128, and #191 (among probably other places), we don't want to have two different ways to do similar things unless there is a very good reason.

kohenkatz (commented):
I'm not sure this is better. For those of us who understand the construction of `time.Time` values from Unix timestamps, the meaning is clear. 

But I think that anyone who does not understand that `time.Unix()` uses the `Local` time zone might think that this means that we are counting the Unix timestamp from the local time zone's `1970-01-01 00:00:00` instead of from UTC.

How about this?

> // Time returns the time.Time representation of a Timestamp.
> //
> // For display and comparison purposes, note that this function produces a time.Time value that contains only wall clock time (no monotonic counter) and uses the time.Local time zone for display.

It simplifies the basic description, and puts the details further down for anyone who cares.

nathanmcgarvey-modopayments:
Counter offer (slightly reworded and formatted):

```
// Time returns the time.Time representation of a Timestamp.
//
// The resulting time.Time:
// - Contains only wall clock time (no monotonic clock component)
// - Uses the local system timezone for the location
```

kohenkatz:
> Counter offer (slightly reworded and formatted):

That looks good to me

kohenkatz (approved):
LGTM

PR #199, opened 2025-02-09
Proper timestamp calculation for UUID v7

### LINKED ISSUE #195: TimestampFromUUIDv7 Artificially Limited

kohenkatz:
Fixes #195.

The use of the `time.Time` method `UnixNanos()` was causing an `int64` overflow (officially an "undefined behavior"), and returning incorrect timestamp values.

There wasn't any good reason for using this function, since the goal is to convert milliseconds to 100-nanosecond intervals, which can be done much more simply with a single multiplication step.

---

This might also provide a slight performance improvement (see https://github.com/gofrs/uuid/issues/128#issuecomment-2106517809), since `time.UnixMilli`, the (unneeded) call to `.UTC()`, and `.UnixNano()` are each doing a fair bit of work internally, and they have all been replaced with a single multiplication.

### DIFF
--- a/generator_test.go
+++ b/generator_test.go
@@ -1101,7 +1101,6 @@ func TestDefaultHWAddrFunc(t *testing.T) {
 }
 
 func BenchmarkGenerator(b *testing.B) {
-	b.ReportAllocs()
 	b.Run("NewV1", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
 			NewV1()
--- a/generator_test.go
+++ b/generator_test.go
@@ -1101,6 +1101,7 @@ func TestDefaultHWAddrFunc(t *testing.T) {
 }
 
 func BenchmarkGenerator(b *testing.B) {
+	b.ReportAllocs()
 	b.Run("NewV1", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
 			NewV1()
@@ -1119,6 +1120,16 @@ func BenchmarkGenerator(b *testing.B) {
 	b.Run("NewV5", func(b *testing.B) {
 		for i := 0; i < b.N; i++ {
 			NewV5(NamespaceDNS, "www.example.com")
+		}
+	})
+	b.Run("NewV6", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			NewV6()
+		}
+	})
+	b.Run("NewV7", func(b *testing.B) {
+		for i := 0; i < b.N; i++ {
+			NewV7()
 		}
 	})
 }
--- a/uuid.go
+++ b/uuid.go
@@ -147,10 +147,10 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 		(int64(u[4]) << 8) |
 		int64(u[5])
 
-	// UUIDv7 stores MS since 1979-01-01 00:00:00, but the Timestamp
+	// UUIDv7 stores MS since 1970-01-01 00:00:00, but the Timestamp
 	// type stores 100-nanosecond increments since 1582-10-15 00:00:00.
 	// This conversion multiplies ms by 10,000 to get 100-ns chunks and adds
-	// the difference between October 1582 and January 1979.
+	// the difference between October 1582 and January 1970.
 	tsNanos := epochStart + (t * _100nsPerMillisecond)
 	return Timestamp(tsNanos), nil
 }
--- a/uuid.go
+++ b/uuid.go
@@ -89,7 +89,6 @@ const (
 type Timestamp uint64
 
 const _100nsPerSecond = 10000000
-const _100nsPerMillisecond = 10000
 
 // Time returns the time.Time representation of a Timestamp.
 //
@@ -151,7 +150,7 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 	// type stores 100-nanosecond increments since 1582-10-15 00:00:00.
 	// This conversion multiplies ms by 10,000 to get 100-ns chunks and adds
 	// the difference between October 1582 and January 1970.
-	tsNanos := epochStart + (t * _100nsPerMillisecond)
+	tsNanos := epochStart + (t * 10000)
 	return Timestamp(tsNanos), nil
 }
 
--- a/uuid.go
+++ b/uuid.go
@@ -82,8 +82,10 @@ const (
 )
 
 // Timestamp is the count of 100-nanosecond intervals since 00:00:00.00,
-// 15 October 1582 within a V1 UUID. This type has no meaning for other
-// UUID versions since they don't have an embedded timestamp.
+// 15 October 1582 within a V1 or V6 UUID, or as a common intermediate
+// representation of the (Unix Millisecond) timestamp within a V7 UUID.
+// This type has no meaning for other UUID versions since they don't
+// have an embedded timestamp.
 type Timestamp uint64
 
 const _100nsPerSecond = 10000000
--- a/uuid.go
+++ b/uuid.go
@@ -87,6 +87,7 @@ const (
 type Timestamp uint64
 
 const _100nsPerSecond = 10000000
+const _100nsPerMillisecond = 10000
 
 // Time returns the time.Time representation of a Timestamp.
 //
@@ -145,7 +146,7 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 		int64(u[5])
 
 	// convert to format expected by Timestamp
-	tsNanos := epochStart + time.UnixMilli(t).UTC().UnixNano()/100
+	tsNanos := epochStart + (t * _100nsPerMillisecond)
 	return Timestamp(tsNanos), nil
 }
 
--- a/uuid.go
+++ b/uuid.go
@@ -147,7 +147,10 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 		(int64(u[4]) << 8) |
 		int64(u[5])
 
-	// convert to format expected by Timestamp
+	// UUIDv7 stores MS since 1979-01-01 00:00:00, but the Timestamp
+	// type stores 100-nanosecond increments since 1582-10-15 00:00:00.
+	// This conversion multiplies ms by 10,000 to get 100-ns chunks and adds
+	// the difference between October 1582 and January 1979.
 	tsNanos := epochStart + (t * _100nsPerMillisecond)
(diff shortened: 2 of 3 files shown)


uuid_test.go
```diff
@@ -262,11 +262,11 @@ func TestTimestampFromV7(t *testing.T) {
 		want    Timestamp
 		wanterr bool
 	}{
-		{u: Must(NewV1()), wanterr: true},
```
nathanmcgarvey-modopayments, line 265: Did you mean to remove this test case that throws an error?

nathanmcgarvey-modopayments (commented):
Would you potentially accept a couple of benchmarks if I wrote them?

uuid_test.go
kohenkatz: That test case relevant here. It was a copy paste mistake from the v1 tests. In V7, it doesn't return an error.

kohenkatz:
> Would you potentially accept a couple of benchmarks if I wrote them?

Sure!

nathanmcgarvey-modopayments:
> > Would you potentially accept a couple of benchmarks if I wrote them?
> 
> Sure!

https://github.com/kohenkatz/uuid/pull/1

uuid_test.go
nathanmcgarvey-modopayments: That's surprising since it is in the v1 and v6 tests and there is an actual case statment to handle it:

```golang
		switch {
		case tt.wanterr && err == nil:
			t.Errorf("TimestampFromV7(%v) want error, got %v", tt.u, got)
```

Surely it should be an error if you try to do a TimestampFromV7 with a NewV1 UUID.

kohenkatz: @nathanmcgarvey-modopayments You are right, I was misreading it. I am putting it back with a note explaining why it's there.

nathanmcgarvey-modopayments (approved):

generator_test.go
```diff
@@ -1101,6 +1101,7 @@ func TestDefaultHWAddrFunc(t *testing.T) {
 }
 
 func BenchmarkGenerator(b *testing.B) {
+	b.ReportAllocs()
```
dylan-bourque, line 1104: Personal opinion: it's cleaner to not have this directly in the benchmark and include `-benchmem` when running `go test`.  This way you **can't** only benchmark times without modifying the code.

uuid.go
```diff
@@ -141,7 +144,7 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 		int64(u[5])
 
 	// convert to format expected by Timestamp
-	tsNanos := epochStart + time.UnixMilli(t).UTC().UnixNano()/100
+	tsNanos := epochStart + (t * _100nsPerMillisecond)
```
dylan-bourque, line 147: I'd like to see an explanatory comment here since now it's "magic math".

uuid_test.go
```diff
@@ -281,6 +284,56 @@ func TestTimestampFromV7(t *testing.T) {
 	}
 }
 
+func TestTimestampMinMaxV167(t *testing.T) {
```
dylan-bourque, line 287:
```suggestion
func TestMinMaxTimestamps(t *testing.T) {
```
Timestamps are only valid for V1, V6 and V7 values and mashing the 3 version together makes for a clunky name IMO.

uuid_test.go
```diff
@@ -300,3 +353,94 @@ func BenchmarkFormat(b *testing.B) {
 		})
 	}
 }
+
+var uuidBenchmarkSink UUID
+var timestampBenchmarkSink Timestamp
+var timeBenchmarkSink time.Time
+
+func BenchmarkTimestampFrom(b *testing.B) {
+	b.ReportAllocs()
```
dylan-bourque, line 362: Same note here re: forcing memory benchmarking

uuid_test.go
```diff
@@ -300,3 +353,94 @@ func BenchmarkFormat(b *testing.B) {
...
+				if err != nil {
+					b.Fatal(err)
+				}
+			}
+		})
+	}
+}
+
+func BenchmarkTimestampTime(b *testing.B) {
+	b.ReportAllocs()
```
dylan-bourque, line 403: And here

kohenkatz:
@dylan-bourque Updated with your suggestions.

uuid.go
```diff
@@ -144,8 +147,11 @@ func TimestampFromV7(u UUID) (Timestamp, error) {
 		(int64(u[4]) << 8) |
 		int64(u[5])
 
-	// convert to format expected by Timestamp
-	tsNanos := epochStart + time.UnixMilli(t).UTC().UnixNano()/100
+	// UUIDv7 stores MS since 1979-01-01 00:00:00, but the Timestamp
+	// type stores 100-nanosecond increments since 1582-10-15 00:00:00.
+	// This conversion multiplies ms by 10,000 to get 100-ns chunks and adds
+	// the difference between October 1582 and January 1979.
```
nathanmcgarvey-modopayments, line 153: s/1979/1970

nathanmcgarvey-modopayments (changes requested):

uuid.go
nathanmcgarvey-modopayments: (two occurrences)

kohenkatz: Fixed. Must have been sitting a little to far to the left relative to my keyboard... 😅

nathanmcgarvey-modopayments (approved):

uuid.go
```diff
@@ -82,11 +82,14 @@ const (
...
-// 15 October 1582 within a V1 UUID. This type has no meaning for other
-// UUID versions since they don't have an embedded timestamp.
+// 15 October 1582 within a V1 or V6 UUID, or as a common intermediate
+// representation of the (Unix Millisecond) timestamp within a V7 UUID.
+// This type has no meaning for other UUID versions since they don't
+// have an embedded timestamp.
 type Timestamp uint64
 
 const _100nsPerSecond = 10000000
+const _100nsPerMillisecond = 10000
```
dylan-bourque, line 92: One more nit: not really a fan of defining a named constant that's only used in a single place as an implementation detail.  it breaks code locality.

kohenkatz: I thought about that, but I'd rather have both time conversion constants next to each other so future devs can see the relationship between them.

dylan-bourque: counter-point: the name `_100nsPerMillisecond` doesn't convey what the value represents (there aren't 100 ns in a millisecond 🤨) or how it's used.  a casual reader would see this and have to grep the code to figure out why this constant exists.

fwiw I have the same gripe with `_100nsPerSecond` but that one precedes my being a maintainer.

kohenkatz: I it's not worth spending more time discussing this. Since there's already a comment that explains the conversion, I changed it to just use a "magic number" there instead of a constant.

dylan-bourque: > I understand the names to perfectly convey what the value represents 

in a way this reinforces my gripe.  we both read the same code and interpreted the name differently. 😉

dylan-bourque (approved):

Issue #48, opened 2018-08-27
SQL value more efficient as bytes


ricardofandrade:
The SQL [`Value()`](https://github.com/gofrs/uuid/blob/d41eeda0759468834c84365a32d1b1ec4264c75f/sql.go#L32) gives a string representation of UUID when called. However, storing UUIDs as bytes may be more efficient for some [databases](http://mysqlserverteam.com/storing-uuid-values-in-mysql-tables/).
Other UUID libraries have the same issue: https://github.com/google/uuid/issues/20

Would it be possible to have an option to either set globally to output bytes or a similar approach used for `NullUUID`?

I'd be glad to work on a patch in either direction if necessary.

zerkms:
I think one possible implementation would be to create a wrapper type:

1. Whose constructor accepts `uuid.UUID`
2. Which implements `driver.Valuer` that returns `[]byte`

In that case it does not incur a BC, and follows the common sense design practices.

A quick and dirty API sample (with deliberately verbose declarations):

```go
var u UUID = ...

sqlByteUuid := NewSQLByteUUID(u)

var b []byte
var err error

b, err = sqlByteUuid.Value()
```

ricardofandrade:
@zerkms thanks for the quick reply.
Since two SQL type conversions are already included (`UUID` and `NullUUID`), would you propose this wrapper type to also be included in `uuid`?

ricardofandrade:
I am currently experimenting with your design.I may propose a change if the ergonomics of using it within `uuid` are better.

theckman:
I think it's solvable in a backwards compatible way without _another_ type.

ricardofandrade:
Are you proposing extending `NullUUID` @theckman or `UUID` itself?

theckman:
I'm naively thinking of extending the NullUUID type to have a mode, which allows you to use either a string or raw bytes mode. GopherCon has me a little distracted, so I'll admit I'm to the level of spitballing while not having thought it through fully. 😊

The one part I've not thought of is pulling the data from the DB. We could maybe use the data length to know if it's a set of bytes, or a string itself.

ricardofandrade:
When pulling from the database what @theckman  said is already being done:
https://github.com/gofrs/uuid/blob/aeac6c4a7786adf624dc27d38504c9011b1ffd58/sql.go#L40

The problem is that when you write back to the database only string is returned:
[...]

PR #235, opened 2025-12-11
Add package flag to change Value() return type to []byte

PauliusLozys:
This adds a package level toggle `SQLEnableBinaryValue` to change the behavior of `Value()` method.
Currently, the `Value()` implementation returns `String()` which causes problems when saving `UUID` to the database (e.g. MariaDB) which uses `BINARY(16)` for `UUID` representation. 
Example code:
```go
	_, err = db.Exec("CREATE TABLE IF NOT EXISTS test_uuid (uid BINARY(16) PRIMARY KEY)")
	if err != nil {
		panic(err)
	}

	uid, err := uuid.NewV7()
	if err != nil {
		panic(err)
	}

	_, err = db.Exec("INSERT INTO test_uuid (uid) VALUES (?)", uid)
	if err != nil {
		panic(err)
	}
```
```
panic: Error 1406 (22001): Data too long for column 'uid' at row 1
```

To circumvent this, the only way is to convert `UUID` to `[]byte` slice using `Bytes()` before passing it as an argument.
```go
	_, err = db.Exec("INSERT INTO test_uuid (uid) VALUES (?)", uid.Bytes())
	if err != nil {
		panic(err)
	}
```
This makes it easy to accidentally pass unconverted value which will later call `Value()`, turn it into a string, and cause an error.
This can also cause subtle bugs in `WHERE ... IN (...)` statements, where unconverted `UUID` will not match any row even if it exists.
```go
	_, err = db.Exec("INSERT INTO test_uuid (uid) VALUES (?)", uid.Bytes())
	if err != nil {
		panic(err)
	}

	rows, err := db.Query("SELECT * FROM test_uuid WHERE uid IN (?)", uid)
	if err != nil {
		panic(err)
	}
	defer rows.Close()

	var uuids []uuid.UUID
	for rows.Next() {
		var uid uuid.UUID
		if err := rows.Scan(&uid); err != nil {
			panic(err)
		}
		uuids = append(uuids, uid)
	}

	fmt.Println("Found UUIDs:", uuids)
```
```
Found UUIDs: []
```
The cleanest and most ergonomic way to solve these problem would be if it was possible to toggle in the package what type is returned from `Value()` method. Then, no conversions would be needed and `UUID` could be passed to queries as is.

### DIFF
--- a/sql.go
+++ b/sql.go
@@ -30,8 +30,22 @@ import (
 var _ driver.Valuer = UUID{}
 var _ sql.Scanner = (*UUID)(nil)
 
+// SQLEnableBinaryValue controls the format of the UUID returned by
+// UUID.Value().
+// By default, UUID.Value() returns the UUID as a string.
+// If SQLEnableBinaryValue is set to true, UUID.Value() returns
+// the UUID as a 16-byte slice.
+var SQLEnableBinaryValue = false
+
 // Value implements the driver.Valuer interface.
+// By default, it returns the UUID as a string.
+// If the package variable SQLEnableBinaryValue is set to true,
+// it returns the UUID as a 16-byte slice.
 func (u UUID) Value() (driver.Value, error) {
+	if SQLEnableBinaryValue {
+		return u.Bytes(), nil
+	}
+
 	return u.String(), nil
 }
 
--- a/sql_test.go
+++ b/sql_test.go
@@ -22,13 +22,17 @@
 package uuid
 
 import (
+	"bytes"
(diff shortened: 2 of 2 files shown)


kohenkatz:
> Unfortunately, the codebases I'm working with were made before `UUID` type was existed and `BINARY(16)` was used instead.

That's an important additional piece of information.

I don't think supporting non-native database types is in-scope for this library.

See #48 which provides some discussion of workarounds that may help you.

PR #241, opened 2026-03-02
Feature/uuidv8

cameracker:
This PR adds v8 support to the package.

This variant is for "custom" implementations where there is very little standardized other than the the bit arrangement of the UUID. Version and Variant bits are set same as other UUID formats.

We will need to release this as a major version update because it adds a function to the `Generator` interface.

EDIT: Reviewing a previous decision about adding vNAtTime() to the Generator interface without incrementing a major version, I'd be comfortable with going the same route. Open to feedback.

### DIFF
--- a/generator.go
+++ b/generator.go
@@ -436,13 +436,13 @@ func (g *Gen) NewV8(customA []byte, customB []byte, customC []byte) (UUID, error
 
 	// Validate input lengths
 	if len(customA) != 6 {
-		return Nil, fmt.Errorf("%w: customA must be exactly 6 bytes, got %d", ErrV8FieldLength, len(customA))
+		return Nil, fmt.Errorf("%w: customA must be exactly 6 bytes (48 bits), got %d", ErrV8FieldLength, len(customA))
 	}
 	if len(customB) != 2 {
-		return Nil, fmt.Errorf("%w: customB must be exactly 2 bytes, got %d", ErrV8FieldLength, len(customB))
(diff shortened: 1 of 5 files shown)


generator.go
```diff
@@ -403,6 +417,52 @@ func (g *Gen) NewV7AtTime(atTime time.Time) (UUID, error) {
...
+	copy(u[0:6], customA)
+
+	// Copy customB (12 bits from 2 bytes) into u[6:8]
+	u[6] = customB[0] & 0x0f  // mask high 4 bits (version goes here)
+	u[7] = customB[1]
+
+	// Copy customC (62 bits from 8 bytes) into u[8:16]
+	// The high 2 bits of u[8] will be overwritten by variant
+	copy(u[8:16], customC)
+	u[8] &= 0x3f // mask off high 2 bits that will be used for variant
```
kohenkatz: The `SetVariant` method already includes `&(0xff>>2)`, which send to do the same thing. Is there a reason we need to do it twice?

cameracker: This is addressed, thank you :)

kohenkatz (dismissed):
LGTM

cameracker:
> A consideration: #246 could also pedantically be considered a breaking change (since existing consumers using Go < 1.25 will stop working). I would argue against a major version for that change as well, but maybe it's worth it if we have more than one breakage. 🤔

I'll put up an issue and ask for community feedback for a couple of weeks and then kick it out as a minor if we dont hear anything

cameracker:
Alight I'm gonna merge this thing in next but git decided the readme had a conflict for some reason.... Needs a reapproval

dylan-bourque (approved):

PR #247, opened 2026-03-17
perf: Improve error handling performance by removing expensive formatting

AmritM18:
Replace fmt.Errorf calls with lightweight error literals to improve parsing performance, following the same approach as Google's UUID library.

We noticed a difference in performance between the Google uuid library and gofrs for uuid parse failures. Wanted to raise this diff to ask why not just return simpler error messages.

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,5 +1,7 @@
 package uuid
 
+import "fmt"
+
 // Error is a custom error type for UUID-related errors
 type Error string
 
@@ -32,6 +34,19 @@ const (
 
 	// ErrInvalidVersion indicates an unsupported or invalid UUID version.
 	ErrInvalidVersion = Error("uuid:")
+)
+
+// Wrapped errors for backward compatibility. These wrap ErrIncorrectFormatInString
+// so code like errors.Is(err, uuid.ErrIncorrectFormatInString) continues to work.
+var (
+	// ErrInvalidBraces is returned when braced format has invalid braces.
+	ErrInvalidBraces = fmt.Errorf("%w: invalid braces", ErrIncorrectFormatInString)
+
+	// ErrInvalidURNPrefix is returned when URN format has invalid prefix.
+	ErrInvalidURNPrefix = fmt.Errorf("%w: invalid URN prefix", ErrIncorrectFormatInString)
+
+	// ErrInvalidDashes is returned when canonical format has dashes not in expected positions.
+	ErrInvalidDashes = fmt.Errorf("%w: dashes were not in expected positions", ErrIncorrectFormatInString)
 )
 
 // Error returns the string representation of the UUID error.
--- a/codec.go
+++ b/codec.go
@@ -21,8 +21,6 @@
 
 package uuid
 
-import "fmt"
-
 // FromBytes returns a UUID generated from the raw byte slice input.
 // It will return an error if the slice isn't 16 bytes long.
 func FromBytes(input []byte) (UUID, error) {
@@ -73,22 +71,22 @@ func parseBytes(b []byte, u *UUID) error {
 	case 36: // canonical
 	case 34, 38:
 		if b[0] != '{' || b[len(b)-1] != '}' {
-			return fmt.Errorf("%w %q", ErrIncorrectFormatInString, b)
+			return ErrInvalidBraces
 		}
 		b = b[1 : len(b)-1]
 	case 41, 45:
 		if string(b[:9]) != "urn:uuid:" {
-			return fmt.Errorf("%w %q", ErrIncorrectFormatInString, b[:9])
+			return ErrInvalidURNPrefix
 		}
 		b = b[9:]
 	default:
-		return fmt.Errorf("%w %d in string %q", ErrIncorrectLength, len(b), b)
+		return ErrIncorrectLength
 	}
 
 	// canonical (36 chars with dashes at fixed positions)
 	if len(b) == 36 {
 		if b[8] != '-' || b[13] != '-' || b[18] != '-' || b[23] != '-' {
-			return fmt.Errorf("%w %q", ErrIncorrectFormatInString, b)
+			return ErrInvalidDashes
 		}
 		for i, x := range [16]byte{
 			0, 2, 4, 6,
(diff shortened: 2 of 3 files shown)


dylan-bourque:
> why not just return simpler error messages

The previous error messages were more informative since we intentionally include additional information about what was actually wrong instead of a single hard-coded string.

AmritM18:
> > why not just return simpler error messages
> 
> The previous error messages were more informative since we intentionally include additional information about what was actually wrong instead of a single hard-coded string.

Should this be a responsibility of the library though? Users that want detailed error context can add it themselves at the call site, no? 

This change would actually bring it in line with Google's approach and help performance quite a bit. Here are some benchmarks where I tested valid and invalid UUIDs. 

```
goos: linux
goarch: amd64
cpu: AMD EPYC 7B13
BenchmarkUUIDLibraries/Valid_Gofrs-48           30231711                38.58 ns/op            0 B/op          0 allocs/op
BenchmarkUUIDLibraries/Valid_Google-48          29218924                39.60 ns/op            0 B/op          0 allocs/op
--
BenchmarkUUIDLibraries/Invalid_Gofrs-48          3398185               363.0 ns/op            96 B/op          2 allocs/op
BenchmarkUUIDLibraries/Invalid_Google-48        86891918                14.92 ns/op            4 B/op          0 allocs/op
--
BenchmarkUUIDLibraries/Mixed_Gofrs-48            5753064               208.6 ns/op            54 B/op          1 allocs/op
BenchmarkUUIDLibraries/Mixed_Google-48          63334467                18.47 ns/op            0 B/op          0 allocs/op
```

The error formatting brings in an unhealthy amount of overhead (24x slower).

dylan-bourque:
You make a valid point about the cost.  The idea was to make the error messages as useful as possible, but you're not wrong in saying that the callers can do that.

A caveat would be that this would _technically_ be a breaking change since there could be existing consumers that are expecting those specific error strings. 😬

@cameracker @kohenkatz WDYT?

zerkms:
Different types of errors can play as a middle ground: returning `ErrInvalidFormat` for every reason is too vague.

AmritM18:
> Different types of errors can play as a middle ground: returning `ErrInvalidFormat` for every reason is too vague.

Updated to have some more types of errors instead of just `ErrInvalidFormat` for everything. Let me know if anything should be added or omitted.

zerkms:
Hold on, I'm just another random person that watches this repository, please don't take my thoughts out loud as instructions :-)

kohenkatz:
I agree with @zerkms - we don't necessarily need to have fully dynamic errors, but we should not make the errors too much less informative.

However, we should also not let this be a breaking change. @dylan-bourque I am not concerned about the specific string contents; rather the ability to use `errors.Is`: until now, many of these errors would be caught by `if errors.Is(err, uuid.ErrIncorrectFormatInString)`, but will no longer be caught by that. There's a simple fix for that, I think - make the new errors wrap the previous error, like this:

```
	// ErrInvalidBraces is returned when braced format has invalid brackets.
	ErrInvalidBraces = fmt.Errorf("%w: invalid braces", ErrIncorrectFormatInString)

	// ErrInvalidURNPrefix is returned when URN format has invalid prefix.
	ErrInvalidURNPrefix = fmt.Errorf("%w: invalid URN prefix", ErrIncorrectFormatInString)

	// ErrInvalidDashes is returned when canonical format has missing/misplaced dashes.
	ErrInvalidDashes = fmt.Errorf("%w: invalid dashes", ErrIncorrectFormatInString)
```

I think that should be a good compromise between backward compatibility and performance, since each of these three wrapped errors is only created once.

error.go
```diff
@@ -32,6 +32,15 @@ const (
...
 	ErrInvalidVersion = Error("uuid:")
+
+	// ErrInvalidBraces is returned when braced format has invalid brackets.
+	ErrInvalidBraces = Error("uuid: invalid braces")
+
+	// ErrInvalidURNPrefix is returned when URN format has invalid prefix.
+	ErrInvalidURNPrefix = Error("uuid: invalid URN prefix")
+
+	// ErrInvalidDashes is returned when canonical format has missing/misplaced dashes.
+	ErrInvalidDashes = Error("uuid: invalid dashes")
```
cameracker:
```suggestion
	ErrInvalidDashes = Error("uuid: dashes were not in expected positions")
```

error.go
```diff
@@ -32,6 +32,15 @@ const (
 
 	// ErrInvalidVersion indicates an unsupported or invalid UUID version.
 	ErrInvalidVersion = Error("uuid:")
+
+	// ErrInvalidBraces is returned when braced format has invalid brackets.
+	ErrInvalidBraces = Error("uuid: invalid braces")
```
cameracker: Giga nit - lets be consistent with the noun we use - lets switch it to braces consistently (there are comments here and in the test file as well)

kohenkatz: Technically, `{}` are braces. Brackets are `[]` (or possibly also `<>`).

cameracker:
I'm neutral. I understand that the 

> You make a valid point about the cost. The idea was to make the error messages as useful as possible, but you're not wrong in saying that the callers can do that.
> 
> A caveat would be that this would _technically_ be a breaking change since there could be existing consumers that are expecting those specific error strings. 😬
> 
> @cameracker @kohenkatz WDYT?

Personally, I'm ambivalent. Perhaps careless, but I'm not as worried about a little perf hit on error handling as I would be about the happy path parsing case. That said, this proposal does seem like a net improvement while still providing informative error messages, so I'm thankful for the contribution :)

error.go
cameracker: Ha, yes, thank you 🤦

AmritM18:
Updated to incorporate feedback, thanks for the prompt reviews everyone :)

dylan-bourque:
👍 from me after the latest changes. Looks like you need to sync with changes on `main`, though.

Ping when it's ready for final review.

AmritM18:
Synced with main. Should be ready for final review @dylan-bourque.

dylan-bourque (approved):
🚢

cameracker (approved):

PR #257, opened 2026-08-03
Fix integer overflow on 32bit architectures

gibmat:
Explicitly create `atMs` as an `int64`, otherwise on 32bit architectures running the tests fail:

```
# github.com/gofrs/uuid [github.com/gofrs/uuid.test]
src/github.com/gofrs/uuid/generator_test.go:1169:97: cannot use atMs (untyped int constant 1645557742000) as int value in argument to t.Errorf (overflows)
```


dylan-bourque (approved):
Good catch! Thanks for the fix.

cameracker:
@gibmat  thanks very much for jumping on this so quickly!

Issue #198, opened 2025-02-09
Should we add checks for valid variant (and version)?


kohenkatz:
> Related, but minor, the tests above, ironically don't set the UUIDv7 variant bits correctly. (the relevant hex character should probably be a `b` for the max UUIDv7 and an `8` for the min UUDv7)  

 _Originally posted by @nathanmcgarvey-modopayments in [#195](https://github.com/gofrs/uuid/issues/195#issuecomment-2645921765)_

Looking through our tests, we have a number of places where a text-format UUID ends with `ffff-ffffffffffff`, even though these should be `bfff-ffffffffffff` to indicate the correct variant.

Should we have a validation in `UUID.Parse` that makes sure the variant is valid? What about making sure the version is valid? Right now, we don't do any checking for this, and we don't have a method for the user to check validity either.

nathanmcgarvey-modopayments:
Variant will be hard to validate as you have all possible combinations of two bits:

```golang
// UUID layout variants.
const (
	VariantNCS byte = iota
	VariantRFC9562
	VariantMicrosoft
	VariantFuture
)
```


So what is "valid" in that field could get squishy unless you want strict RFC conformance. (And at that, I'd have to go look it up to see how strict they are.)

Version, however has to be validated *somewhere*. I can't just get a timestamp from a non-v1,v6,or v7 UUID. But maybe there's no ingest validation. Just when certain parsing is requested.

kohenkatz:
> Version, however has to be validated _somewhere_. I can't just get a timestamp from a non-v1,v6,or v7 UUID. But maybe there's no ingest validation. Just when certain parsing is requested.

Validation for this happens in the `TimestampFromV1`, `TimestampFromV6`, and `TimestampFromV7` methods. For example:

```go
func TimestampFromV7(u UUID) (Timestamp, error) {
	if u.Version() != 7 {
		return 0, fmt.Errorf("%w %s is version %d, not version 7", ErrInvalidVersion, u, u.Version())
	}

	...
```

But there is no validation when creating the UUID object from text (or binary) input.

cameracker:
I was going over the RFC again, and this section covers validation somewhat:

> [6.12. ](https://www.rfc-editor.org/rfc/rfc9562#section-6.12)[Opacity](https://www.rfc-editor.org/rfc/rfc9562#name-opacity)
[...]

PR #261, opened 2026-09-16
Validate timestamp UUID fixture contracts

LeulTew:
Positive timestamp fixtures now check their expected UUID version and RFC 9562 variant independently of timestamp extraction. This corrects five variant fields and prevents min/max fixtures from silently selecting the wrong version-specific extractor; timestamp expectations and production code are unchanged.

Refs #198.

**Validation limits:** Checked on Go 1.27.1, Windows/amd64 only. Race testing is blocked by disabled cgo and no `gcc` on PATH; the Go 1.25/1.26 Linux CI matrix and macOS are unverified. The pinned pre-commit hooks were unavailable, and supplemental golangci-lint 2.13.1 reports three unchanged baseline findings.

<details>
<summary>Implementation and validation</summary>

- Positive `TestTimestampFromV1`, `TestTimestampFromV6`, and `TestTimestampFromV7` fixtures use fixed expected versions and `VariantRFC9562`. The explicit wrong-version cases retain their existing error expectations.
- `TestMinMaxTimestamps` declares each fixture's expected version and uses it for both the assertion and extractor selection, rather than deriving its expected algorithm from the parsed value.
- The original focused tests passed despite five incorrect variants. Adding the assertions first failed on exactly those five fixtures; changing only their variant bits made the tests pass.
- A V1-labelled boundary fixture mutated to V6 passed the original test. The updated assertion rejects it even when the old actual-version dispatch is temporarily restored. Six artifact-only mutation controls exercised the version and variant checks without changing production files.
- All 15 positive timestamp results, time conversions, and expected values were compared before/after and remained identical. All four negative fixtures and every other `FromString` literal remained unchanged.
- Focused tests, full `go test` with atomic coverage, `go build`, and `go vet` passed with one test/package worker. Production statement coverage remained 100.0%, with the same covered blocks.
- The changed file is gofmt-clean. The raw Windows checkout reports ten untouched CRLF files; all canonical baseline Go files are gofmt-clean. No unrelated files were reformatted.
- Supplemental golangci-lint 2.13.1 reports the same unchecked `fmt.Fprintf` calls on the baseline and final source: `uuid.go:268`, `uuid.go:293`, and the unchanged benchmark at `uuid_test.go:394` (baseline line 370). It is not a substitute for the pinned v1.52.2 pre-commit hook.

</details>


uuid_test.go
```diff
@@ -353,6 +367,16 @@ func TestMinMaxTimestamps(t *testing.T) {
 	}
 }
 
+func checkTimestampUUID(t *testing.T, u UUID, version byte) {
```
kohenkatz: Why is this called `checkTimestampUUID` when it has nothing at all to do with the timestamp, rather the version and variant numbers?

uuid_test.go
```diff
@@ -305,28 +314,31 @@ func TestTimestampFromV7(t *testing.T) {
...
-		{u: Must(FromString("ffffffff-ffff-6fff-bfff-ffffffffffff")), want: time.Date(5236, 3, 31, 21, 21, 00, 684697500, time.UTC)}, //5236-03-31 21:21:00 (UTC)
+		{u: Must(FromString("00000000-0000-6000-8000-000000000000")), version: V6, want: time.Date(1582, 10, 15, 0, 0, 0, 0, time.UTC)},           //1582-10-15 0:00:00 (UTC)
+		{u: Must(FromString("ffffffff-ffff-6fff-bfff-ffffffffffff")), version: V6, want: time.Date(5236, 3, 31, 21, 21, 00, 684697500, time.UTC)}, //5236-03-31 21:21:00 (UTC)
 
 		// v7 min and max
-		{u: Must(FromString("00000000-0000-7000-8000-000000000000")), want: time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)},            //1970-01-01 0:00:00 (UTC)
-		{u: Must(FromString("ffffffff-ffff-7fff-bfff-ffffffffffff")), want: time.Date(10889, 8, 2, 5, 31, 50, 655000000, time.UTC)}, //10889-08-02 5:31:50.655 (UTC)
+		{u: Must(FromString("00000000-0000-7000-8000-000000000000")), version: V7, want: time.Date(1970, 1, 1, 0, 0, 0, 0, time.UTC)},            //1970-01-01 0:00:00 (UTC)
+		{u: Must(FromString("ffffffff-ffff-7fff-bfff-ffffffffffff")), version: V7, want: time.Date(10889, 8, 2, 5, 31, 50, 655000000, time.UTC)}, //10889-08-02 5:31:50.655 (UTC)
```
kohenkatz, line 332: Can you better explain the purpose of adding explicit version flags here?

uuid_test.go
```diff
@@ -236,11 +236,14 @@ func TestTimestampFromV1(t *testing.T) {
...
 		{u: Must(NewV4()), wanterr: true},
-		{u: Must(FromString("00000000-0000-1000-0000-000000000000")), want: 0},
+		{u: Must(FromString("00000000-0000-1000-8000-000000000000")), want: 0},
 		{u: Must(FromString("424f137e-a2aa-11e8-98d0-529269fb1459")), want: 137538640775418750},
-		{u: Must(FromString("ffffffff-ffff-1fff-ffff-ffffffffffff")), want: Timestamp(1<<60 - 1)},
+		{u: Must(FromString("ffffffff-ffff-1fff-bfff-ffffffffffff")), want: Timestamp(1<<60 - 1)},
 	}
 	for _, tt := range tests {
+		if !tt.wanterr {
+			checkUUIDVersionAndVariant(t, tt.u, V1)
```
dylan-bourque: Why not restructure the `if`/`else` block below?
```
if tt.wanterr {
  if goterr == nil {
    ...
  }
} else {
  checkUUIDVersionAndVariant(t, tt.u, V1)
  if tt.want != got {
    ...
  }
}
```
I think ☝️ reads clearer than separating out the version/variant check.

The same thing applies to the other calls to `checkUUIDVersionAndVariant()` in the other tests. Don't separate that assertion/check from the others.

uuid_test.go
```diff
@@ -305,28 +314,31 @@ func TestTimestampFromV7(t *testing.T) {
 
 func TestMinMaxTimestamps(t *testing.T) {
 	tests := []struct {
-		u    UUID
-		want time.Time
+		u       UUID
+		version byte
```
dylan-bourque: I would call this `wantVersion` or `expectedVersion` to better indicate what it represents.

uuid_test.go
```diff
@@ -353,6 +367,16 @@ func TestMinMaxTimestamps(t *testing.T) {
 	}
 }
 
+func checkUUIDVersionAndVariant(t *testing.T, u UUID, version byte) {
```
dylan-bourque: Please add a doc comment here that calls out the hard-coded check that variant is `VariantRFC9562`. A casual reader may see this and wonder why there's no `variant` parameter.

